PR #33, opened 2024-12-17
Refactored with notifier pattern

chebizarro:
This PR represents a major refactoring of the App to use the Notifier pattern which makes the application more reactive to user input and interactions with the Mostro daemon and easier to adapt to any changes or additions to the base protocol.

These changes now mean the app supports the following actions:
- Add Buy/Sell orders
- Take Buy/Sell orders
And their respective UX

This is still alpha software and should not be used with production systems. Check the relay config in `lib/app/config.dart` before adding or taking orders to make sure you are not operating on live data. 

Order state is presently not persisted between sessions.

This version still uses the `content` field instead of `payload` and random key assignment. Once MostroP2P/mostro#398 is merged it will be updated accordingly.


grunch (approved):
utACK

Amazing work! now let's move to keys-management 😃

PR #114, opened 2025-06-03
Delete l10n generated files

grunch:
## Summary by CodeRabbit

- **Chores**
  - Updated file ignore settings to exclude files in the lib/generated directory from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,4 +72,4 @@ chrome/.packages
 *.lock
 .pdm
 bfg-1.14.0.jar
-lib/generated/*
+lib/generated/


grunch:
> tACK It works fine. I just have a small concern about `lib/generated/action_localizations.dart`. Since it’s a custom file, ignoring means future changes wouldn’t be tracked but I’m not sure if it might need updates later but if you’re sure it won’t be modified, it is fine to leave it as is

I wanted to ignore the whole directory `lib/generated` but after I deleted `action_localizations.dart` I started to get some error compiling, so I only deleted all `l10n*` files but I'm not sure if is ok to delete other files in `lib/generated/`, about your doubt, since `action_localizations.dart` is already on the project we will have no problem adding `lib/generated/` on the .gitignore file, it will work for new generated files like l10n files.

Still this PR will need an ACK from @chebizarro as mobile dev expert 😃

PR #123, opened 2025-06-19
Bug when selecting the premium or discount for an order

### LINKED ISSUE #120: Bug when selecting the premium or discount for an order
When setting the premium or discount while creating an order, it sometimes gets applied incorrectly. For example, it sends to mostrod the new order  event with a 2% premium when the user selected 3%, and the percentage bar showed 3%.
This depends on how close the user leaves the slider to the actual number.

It seems to me (I haven’t checked the code) that what’s happening is: some time ago, the slider showed numbers with decimals, but the order was sent using only the integer part. For example, if the user selected 2.8%, the new order would be sent with 2%.

I think that after removing the decimals from the slider, this is still happening—except now the user sees rounded integers, and where they see 3, the app is actually taking 2.8 and sending 2.

That’s just my assumption, it should be reviewed.

chebizarro:
Fixes #120

The Slider was being set with a range of 200 units but being rounded up to whole numbers as you suspected. The Slider range is now set to 20, constraining the range of values to {-10 ..10}

## Summary by CodeRabbit

- **Style**
  - Updated the slider's thumb interaction effect for a more refined appearance.

- **Refactor**
  - Reduced the number of slider steps for easier adjustment between minimum and maximum values.

### DIFF
--- a/lib/features/order/widgets/premium_section.dart
+++ b/lib/features/order/widgets/premium_section.dart
@@ -35,7 +35,7 @@ class PremiumSection extends StatelessWidget {
               activeTrackColor: AppTheme.purpleAccent,
               inactiveTrackColor: AppTheme.backgroundInactive,
               thumbColor: AppTheme.textPrimary,
-              overlayColor: AppTheme.purpleAccent.withOpacity(0.2),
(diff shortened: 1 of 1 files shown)


Catrya (approved):
tACK thanks @chebizarro

PR #129, opened 2025-07-02
Make user-created orders also appear in the order book, not just in My Trades

chebizarro:
This PR ensures that orders created by the user are visible in the main order book, not just under "My Trades."

Additionally, the order label now displays a "YOU ARE" indicator next to user-created orders trade type, making it easy to distinguish which orders belong to the current user at a glance. This improves clarity and helps users quickly identify their own orders in the list.


Catrya (commented):
Hi @chebizarro Sometimes when I create an order it doesn't have the label "you are selling" or "you are buying", and when I open the order I get an error.

But if I close and reopen the app it works fine and the error doesn't appear.

Captura desde 2025-07-03 09-51-03

Catrya:
Esta primera orden que aparece la creé yo, pero no me sale la etiqueta por mí. Si la orden es tomada, sí me avisa que alguien la tomó y la puedo completar.
Captura desde 2025-07-08 13-24-15
Captura desde 2025-07-08 13-24-21

grunch:
ping @chebizarro

grunch (approved):
tACK

Issue #67, opened 2025-04-29
Create a button to change the language


Catrya:
There should be a button for users to change the app's language, or for the app to use the phone's default language.

PR #138, opened 2025-07-08
Create a button to change the language

chebizarro:
Implements #67
- Added a language selector to the app settings, allowing users to override the app language or use the device default.
- Integrated all relevant localization strings and resolved ARB file issues.
- The app now updates its language immediately based on user selection or falls back to the system locale.

## Summary by CodeRabbit

* **New Features**
  * Added a language selection option in the settings, allowing users to choose their preferred app language or use the system default.
  * Introduced a language selector widget with support for English, Spanish, and Italian.
  * Updated the settings screen with a new language section.
  * Improved locale handling to dynamically prioritize user-selected language or system locale with a refined fallback mechanism.

* **Localization**
  * Added new localization strings for language selection in English, Spanish, and Italian.

* **Bug Fixes**
  * The app now correctly uses the user-selected language preference instead of always defaulting to Spanish when the system locale is Spanish.

### DIFF
--- a/lib/core/app.dart
+++ b/lib/core/app.dart
@@ -54,8 +54,7 @@ class _MostroAppState extends ConsumerState<MostroApp> {
           theme: AppTheme.theme,
           darkTheme: AppTheme.theme,
           routerConfig: goRouter,
-          // Force Spanish locale for testing if device is Spanish
-          locale: systemLocale.languageCode == 'es' ? const Locale('es') : null,
+          // Let localeResolutionCallback handle all locale detection
           localizationsDelegates: const [
             S.delegate,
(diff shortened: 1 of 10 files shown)


grunch (changes requested):
I just installed the app on this branch and it starts on english while my mobile is on spanish, it is important to let you know that right now, in main, the app starts on the system default language (spanish in my case), we need to keep this behavior

Catrya (commented):
Now it does appear in the default language. But the language is selected directly and it doesn't allow selecting the default language afterward.
image

Catrya (approved):
tACK @chebizarro @grunch

PR #153, opened 2025-07-16
change user-npub for user-nym and improve messages

### LINKED ISSUE #58: Improve the message to contact the counterpart
In Active's message It should say that contact the counterpart playing in the contact button.
Also that does not say the trade key, but the NYM and when he touched it the chat opens with the counterpart
Image

Catrya:
fix #58 
Now, the counterparty's npub is not displayed, but rather the nym, highlighted in bold for a better UI.
Additionally, the wording of the active, fiatsent, and payment completion status messages has been improved in all three languages for a better UX.
image


grunch (approved):
utACK

PR #241, opened 2025-07-25
feat: add basic support for PaymentFailure and NextTrade payload types

chebizarro:
This PR adds basic support for the PaymentFailure and NextTrade payload types recently added to the mostro protocol

Users notifications and the Order Details views will now display the number of retry attempts correctly in the event of a PaymentFailure

These changes should be backwards compatible with earlier mostrod versions

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying detailed information when a payment fails, including the number of attempts and retry intervals.
  * Introduced new message types for handling upcoming trades and failed payments.
  * Added a new "Payment Failed" status with corresponding actions for buyers and sellers.

* **Improvements**
  * Payment failure messages now show actual attempt and retry values instead of placeholders, providing more accurate and informative feedback to users.
  * Updated order status indicators and transitions to reflect payment failure state.
  * Action buttons are now suppressed during payment failure state, showing only the Close button to avoid confusion.
  * Enhanced UI padding to better accommodate device safe areas across multiple screens.
  * Refined dialog button layouts and spacing for a cleaner and more consistent user interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,33 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.0+4]
+
+### Added
+- Comprehensive relay URL validation with proper WebSocket protocol checking
+- Real-time relay connectivity testing using direct WebSocket connections
+- Loading indicators during relay validation and testing process
+- Enhanced error messages for invalid relay URLs with helpful formatting hints
+- Debug-only display mode for Current Trade Index Card (hidden in release builds)
+
+### Fixed
+- Relay connectivity testing now accurately detects non-existent or unreachable relays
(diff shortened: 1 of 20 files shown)


Catrya (commented):
@chebizarro 
There are two parts:
1- When the first payment fails and Mostro sends the payment-failed message (it's still trying to pay more times but isn't asking for a new invoice).
2- When all payment attempts have failed and Mostro sends the `add-invoice` action, but with the status `settled-hold-invoice`.

1. In the first part, only the payment failure message and the close button should appear. The user must wait for Mostro to complete the payment or request a new invoice. The Dispute, Add Invoice, and Cancel buttons should not appear.
Also, mostro sends `payment_retries_interval` in seconds, so it must be converted to minutes to show the user the time in minutes (in this case, payment_retries_interval=60 sec, but it shows 60 minutes because it hasn't been converted).

image  


2. In the second part, mostro sends the addinvoice message. In My Trades screen, the `payment-failed` status should remain. In Order details screen, the same screen should appear as above, but with only the Close and Add Invoice buttons, but not with the clock or 15 min message
image

Catrya:
Hi @chebizarro  I made some changes to the texts to improve the UX. I've already tested everything and it works fine.

PR #305, opened 2025-09-15
Fix orphan sessions with timeout detection and cleanup

### LINKED ISSUE #64: Bug taking a buy order: the order is listed as Pending in My Trades for the taker, as if they had created it.
It's happening quite randomly (I haven't found the pattern) that when trying to take a buy order, it's as if the app isn't responding. Then, when the taker goes to My Trades, the order appears as Pending, and when they open it, they see the screen as if they were the maker of that order.
This is what mostrod shows in the logs when the taker tried to take the buy order:
```
2025-04-28T19:18:49.236351Z WARN mostrod::app: Signature verification failed: sender 51898b04abfc7ef6667ef45e746a58e59a881042f92cc1a0cddf9a457126d1d6 does not match rumor pubkey 005a46c4884ba212646401200a0f697c9a05d3c1ab1802f4d6ae268a200c6120
```

Catrya:
fix #64 
  - Add 10s timer cleanup for phantom sessions when Mostro doesn't respond
  - Handle maker vs taker scenarios differently (session preservation vs cleanup)
  - Add localized timeout notifications and automatic navigation
  - Update documentation with this new behavior

## Summary by CodeRabbit

* **New Features**
  * Dual orphan-session protection: 10‑second auto-cleanup when taking orders plus real-time public-event timeout detection.
* **UX Improvements**
  * Distinct maker/taker outcomes: makers keep pending orders; takers’ sessions removed and orders return to the book. Timeouts/cancellations surface localized notifications and navigate to the main order book.
* **Localization**
  * Session timeout message added in English, Spanish, Italian.
* **Documentation**
  * Expanded session lifecycle, timer management, cancellation rules; cleanup window extended to 72 hours.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,24 +109,46 @@ The app follows a specific initialization order in `appInitializerProvider`:
 - SubscriptionManager uses `fireImmediately: false` to prevent premature execution
 - Proper sequence ensures orders appear consistently in UI across app restarts
 
-## Timeout Detection & Reversal System
+## Timeout Detection & Orphan Session Prevention
 
 ### Overview
-Real-time system that detects when orders timeout from `waiting-buyer-invoice` or `waiting-payment` states and handles maker/taker scenarios differently to maintain UI consistency with mostrod state.
 
-### Detection Mechanism
+Comprehensive system that prevents orphan sessions and detects order timeouts through dual protection mechanisms: 30-second cleanup timers and real-time timeout detection via public events.
+
+### Orphan Session Prevention
+
+#### **30-Second Cleanup Timer**
+Automatic cleanup system that prevents sessions from becoming orphaned when Mostro instances are unresponsive:
+
+- **Activation**: Started automatically when users take orders (`takeSellOrder`, `takeBuyOrder`)
+- **Purpose**: Prevents orphan sessions when Mostro doesn't respond within 30 seconds
+- **Cleanup**: Deletes session, shows localized notification, navigates to order book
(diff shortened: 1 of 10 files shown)


AndreaDiazCorreia:
> > It looks pretty good. I’m not sure if we should have the timer hardcoded, but other than that it seems fine. One question though: how does the timer work if the user leaves the app in the background?
> 
> Have you tested it? If the app stays in the background, the counter keeps running, and when the user returns to the app, they see the message that there was no response.

Claro pero no se si deberia ser asi porque una persona puede tomar una orden y recibir una llamada por ejemplo, la app queda en background y le sacas la orden por eso, no se si sea el mejor approach en esos casos

Catrya:
> > > It looks pretty good. I’m not sure if we should have the timer hardcoded, but other than that it seems fine. One question though: how does the timer work if the user leaves the app in the background?
> > 
> > 
> > Have you tested it? If the app stays in the background, the counter keeps running, and when the user returns to the app, they see the message that there was no response.
> 
> Claro pero no se si deberia ser asi porque una persona puede tomar una orden y recibir una llamada por ejemplo, la app queda en background y le sacas la orden por eso, no se si sea el mejor approach en esos casos

El timer corre y la sesión se elimina si no hay respuesta de mostro (si el usiario estaba offline y quizo tomar una orden, o si está online y hay error como unwraping gift u otro que hace que no sea exitoso el take order). Pero si llega cualquier mensaje de mostro, como cant-do o addinvoice o cualquier otro, el timer se cancela y continúa con el mensaje que le llegó

PR #309, opened 2025-09-18
feat: implement subscription initialization fix for existing sessions

### LINKED ISSUE #308: When closing and reopening the app, an order in the waiting status doesn't update.
Some orders get stuck in previous statuses. For example, a user creates a sales order (it's created and the order has a Pending status), another user accepts the order, and the order is placed in the waiting-buyer-invoice status. The order creator closes the app (closes it, not puts it in the background). The buyer pays the invoice, and the order moves to the waiting-payment status (it's the seller's turn to pay the invoice). The seller-maker opens the app and doesn't see anything new. The order remains in the UI as waiting-buyer-invoice, but the QR code screen should open, and the order should appear in the waiting-payment status. If the user creates a new order, the stuck order is updated and appears in the waiting-payment status.

However, if the seller closes the app when the order was in Active status, while it is closed the buyer sends fiatsent and the order goes to fiat-sent status when the seller opens the app if the order is updated to fiatsent status correctly.    

It may be related to issue #286 but I'm not completely sure yet.

Catrya:
fix #308

Adds manual initialization to SubscriptionManager to prevent orders from getting stuck in previous states after app restart.

  - Add _initializeExistingSessions() method to SubscriptionManager constructor
  - Preserves fireImmediately: false while ensuring existing sessions get subscriptions
  - Fixes issue where orders appeared stuck in waiting states after app restart
  - Creates dual-path architecture: listener for future changes + manual init for existing data
  - Add comprehensive test to prevent accidental removal of fix
  - Update documentation in APP_INITIALIZATION_ANALYSIS.md and CLAUDE.md

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscriptions for existing sessions are now initialized during app startup so the UI reflects correct states immediately after restart.

* **Documentation**
  * Updated architecture overview and startup initialization flow; refreshed “Last Updated” timestamps.

* **Tests**
  * Added regression/code-verification tests to ensure the new startup initialization path remains intact.

### DIFF
--- a/lib/features/subscriptions/subscription_manager.dart
+++ b/lib/features/subscriptions/subscription_manager.dart
@@ -90,10 +90,9 @@ class SubscriptionManager {
   }
 
   void _updateSubscription(SubscriptionType type, List<Session> sessions) {
-    unsubscribeByType(type);
-
     if (sessions.isEmpty) {
       _logger.i('No sessions for $type subscription');
+      unsubscribeByType(type);
       return;
     }
 
@@ -102,6 +101,8 @@ class SubscriptionManager {
       if (filter == null) {
         return;
       }
+      // Replace existing subscription only when we have a new filter to apply
+      unsubscribeByType(type);
       subscribe(
         type: type,
         filter: filter,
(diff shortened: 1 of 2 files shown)


grunch (approved):
tACK

PR #353, opened 2025-10-30
fix: reset payment methods when fiat currency changes (#352)

### LINKED ISSUE #352: Wrong payment method at fiat change (Creating orders)
When you create a order. If you set a payment method with some fiat and later change the fiat to another one, the payment method will not reset.

For example:
Create order
Set ARS, select "Mercado Pago"
Go back, set USD

The order remember the last payment methods who we are select.
Should be reseting when we change the fiat.

Image

abcb1122:
Fixes #352 by automatically resetting payment methods when fiat currency changes using a Riverpod listener.

## Release Notes

* **Bug Fixes**
  * Fixed payment method state persisting incorrectly when changing currencies. Payment selections and custom payment fields now properly reset when switching to a different currency, ensuring consistent UI behavior.

### DIFF
--- a/lib/features/order/screens/add_order_screen.dart
+++ b/lib/features/order/screens/add_order_screen.dart
@@ -192,6 +192,20 @@ class _AddOrderScreenState extends ConsumerState<AddOrderScreen> {
       ref.watch(exchangeRateProvider(selectedFiatCode));
     }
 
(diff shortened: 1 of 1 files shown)


Catrya (approved):
tACK

PR #390, opened 2025-12-26
Improve Mostro instance configuration UX

### LINKED ISSUE #380: Annoying “Restoring Orders” popup when modifying Mostro key
**Describe the bug**
In Settings, when I try to change the Mostro key by manually deleting it instead of selecting everything, an annoying warning pops up. From a UX perspective, I think it would be better to have a Save button, and when pressing it, the app should check whether there are orders to restore for that Mostro

**To Reproduce**
Go to Settings and try to change the Mostro key by deleting it character by character

**Expected behavior**
Ideally, no warning should appear until the key has the correct number of characters, or—if the length may vary—there should be a Save button, and the app should only check for orders to restore after pressing that button.

**Screenshots**

Image

Image

**Smartphone (please complete the following information):**
 - Device: Motorola
 - OS: Android

Catrya:
fix #380 

  - Add 3-second debounce timer to prevent restore interruptions
  - Implement hex/npub format validation with localized error messages
  - Auto-convert npub format to hex before saving
  - Update info dialog to document supported formats (en/es/it)
  
Now the user can enter the `npub` format in the pubkey field and it will automatically be converted to `hex` format. This does not affect the functionality of the pubkey in `hex` format as it was previously used, it is simply a change to improve the user experience.

## Summary by CodeRabbit

* **New Features**
  * Added input validation for Mostro public key with inline error messages and real-time feedback.
  * Support for both hex and npub formats with automatic conversion and display as hex.

* **Improvements**
  * Debounced processing of public key input to reduce unnecessary updates.
  * Safer update/restore flow with improved error handling and resource cleanup.

* **Localization**
  * Updated help text and added validation message strings in English, Spanish, and Italian.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


grunch (approved):
tACK

BraCR10 (approved):

PR #439, opened 2026-02-10
fix: Prevent combining fixed price with range orders in order creation

### LINKED ISSUE #434: Prevent selecting Fixed Price when creating a range order
Currently, when creating a new order, the app allows selecting a price range and a fixed price, which is incorrect. If the order is submitted this way, Mostrod returns cant-do.

The app should not allow selecting Fixed Price if the user chooses a price range. Additionally, a clear message should be displayed explaining why this option is disabled to avoid confusion.

Image

Catrya:
fix #434 
Now, if someone tries to create a range order at a fixed price, it will not be allowed, and a message will appear for 5 seconds explaining why it remains at market price.
image

## Summary by CodeRabbit

* **New Features**
  * Range mode for orders: specify min/max amounts and toggle between range and fixed-price modes. UI prevents invalid mode switches and shows a temporary error when fixed price isn’t available for a range.
  * Amount input now activates range mode via user interaction for smoother workflow; price control surfaces the range-related error message.

* **Localization**
  * Added/fixed translations: new fixed-price-vs-range message and clearer fiat amount range wording (English, Spanish, Italian).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,18 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 - URL normalization: Trailing slash removal prevents duplicate entries
 - Instance-isolated testing: Test connections don't affect main app connectivity
 
+### Multi-Mostro Instance Support
+- **Overview**: Users can connect to multiple Mostro instances (nodes) from a single app, switching between them with full relay and subscription reset
+- **Node Types**: Trusted nodes (hardcoded in `Config.trustedMostroNodes`, cannot be removed) vs custom nodes (user-added, persisted in SharedPreferences, removable)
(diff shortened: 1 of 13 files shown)


Catrya:
> No esta duplicando la evaluacion de _isRangeMode? veo que en add_order_screen tiene su boolean y que tambien esta en AmountSection, no estaria redundando ahi? de resto lo probe y anda bien, la pregunta es solo curiosidad

No es redundancia, rastrean cosas distintas:  en AmountSection el _isRangeMode indica si el usuario interactuó con el campo max (estado de UI), mientras que en AddOrderScreen refleja si la orden es de rango (_isRangeMode && hasMax), que es lo que llega por el callback.

grunch (approved):
utACK

Issue #276, opened 2025-08-22
Error unwrapping gift


Catrya:
Sometimes (randomly) when trying to create an order, this error appears in the Mostrod logs. 
```
2025-08-22T04:50:17.972653Z  WARN mostrod::app: Error unwrapping gift
```

I don't know if it's because the app sometimes sends a poorly wrapped Gift Wrap or if it's a Mostrod bug.

I've tried to reproduce it but can't. Here are the app logs before sending the new order action, the app waits for a message from Mostrod that doesn't arrive because of that error.

```
│ 💡 Sending DM, Event ID: 206805333442e25d9c223080378beea1ed13980a1b6a413af3ff55cfc07e735b with payload: {version: 1, request_id: 1623314458, trade_index: 2, action: new-order, payload: {order: {kind: sell, status: pending, amount: 0, fiat_code: ARS, min_amount: null, max_amount: null, fiat_amount: 2222, payment_method: Belo, premium: 0, created_at: 0, expires_at: null, buyer_token: null, seller_token: null}}}
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│ #0   NostrService.publishEvent (package:mostro_mobile/services/nostr_service.dart:130:15)
│ #1   MostroService.publishOrder (package:mostro_mobile/services/mostro_service.dart:189:42)
├┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄
│ 💡 Publishing event 206805333442e25d9c223080378beea1ed13980a1b6a413af3ff55cfc07e735b to relays: [wss://relay.mostro.network, wss://nos.lol, wss://relay.nostr.band]
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[...]

PR #496, opened 2026-02-23
Fix BIP-340 signature bug by overriding bip340 to 0.2.0

### LINKED ISSUE #491: Invalid signature error from mostrod when creating orders or restoring user
When using a previously created user (several days old), any interaction with mostrod fails with:

`WARN mostrod::app: Error unwrapping gift: Invalid signature`

This happens when:

• Creating a new order
• Refreshing/updating the user
• Generating a new user and importing it

Steps to reproduce

1. Use the following test mnemonic: `siren total pony spatial fringe submit uniform tape version warm powder blouse`
2. Restore the user in the app (you can't)
3. Try to create any order

Mostrod rejects the message with Invalid signature. No useful logs appear on the mobile side.

App debug info:
`Restore initialized temp trade key with pubkey:2e2c567eebb54b06ea72f2975a468b5ff0124035e39968ce1a46f2812b008e6a`

Related to #276

Catrya:
fix #491 #276

dart_nostr 9.1.1 depends on bip340 0.1.0, which has a bug in bigToBytes() that doesn't pad values to 32 bytes. This causes about 1-2% of Schnorr signatures to be invalid, and 100% failure for keys whose public key starts with 0x00.

The bug was fixed upstream in bip340 0.2.0. We can't go higher than 0.2.0 because version 0.3.0+ changed verify() from String? to String, which breaks dart_nostr's nullable parameter and won't compile.

bip340 0.2.0 has the same API as 0.1.0, just with the padding fix applied. All 218 tests pass.

The override can be removed when dart_nostr updates its bip340 dependency or when we migrate to a different nostr library.

## Summary by CodeRabbit

* **Chores**
  * Updated package dependency configuration to ensure compatibility and proper dependency resolution.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -130,13 +130,13 @@ packages:
     source: hosted
     version: "1.0.0"
   bip340:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       name: bip340
-      sha256: "9a2f454d3f5e2d67a1b21494203132786d95bde8784e7ada74adac00df3ddc11"
+      sha256: "2a92f6ed68959f75d67c9a304c17928b9c9449587d4f75ee68f34152f7f69e87"
       url: "https://pub.dev"
     source: hosted
-    version: "0.1.0"
+    version: "0.2.0"
   bip39:
     dependency: "direct main"
     description:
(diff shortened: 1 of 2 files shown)


AndreaDiazCorreia (approved):
tACK

PR #501, opened 2026-02-25
Phase 2 user-admin chat: unify dispute chat message model with NostrEvent and gift  wrap storage

Catrya:
- Replace DisputeChat model with DisputeChatMessage wrapper around NostrEvent
  - Store gift wrap events (encrypted) on disk, consistent with P2P chat pattern
  - Reconstruct and unwrap gift wraps on historical message load
  - Add getAdminSharedKey() for future multimedia support (Phase 3)
  - Add isFromUser() method to notifier for sender determination
  - Update dispute_messages_list, dispute_content, and read status service
  - Update plan doc: Phase 1 and Phase 2 marked as DONE

## Summary by CodeRabbit

* **New Features**
  * Added message state tracking to display pending and error status for sent messages.
  * Implemented persistent storage for encrypted messages in Dispute Chat.

* **Documentation**
  * Updated Phase 1/Phase 2 transition plan for Dispute Chat feature, clarifying encryption and messaging architecture.

### DIFF
--- a/lib/features/disputes/notifiers/dispute_chat_notifier.dart
+++ b/lib/features/disputes/notifiers/dispute_chat_notifier.dart
@@ -377,18 +377,21 @@ class DisputeChatNotifier extends StateNotifier<DisputeChatState> {
     }
   }
 
-  /// Update a message's pending/error state in the current state
+  /// Update a message's pending/error state in the current state.
+  /// Per-message errors stay at message level; state.error is reserved
+  /// for initialization/loading failures only.
   void _updateMessageState(String messageId, {required bool isPending, String? error}) {
     final updatedMessages = state.messages.map((m) {
       if (m.id == messageId) {
-        return m.copyWith(isPending: isPending, error: error);
+        return DisputeChatMessage(
+          event: m.event,
+          isPending: isPending,
+          error: error,
+        );
       }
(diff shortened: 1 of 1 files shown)


Catrya:
> I tested several disputes, and the expected features planned in this PR are working properly, good job!
> 
> However, I would like to point out some details regarding chat restore handling. After testing, I noticed the following:
> 
> 1. When using the same device to switch between buyer and seller accounts, the app keeps the previous user's chat. This might be happening because the dispute ID remains the same and the chat data is not being cleared.
> 2. Once  users restores their account, there is no possibility to send or receive messages between the admin and the user. The admin does not receive messages, and vice versa.
> 
> I reviewed the implementation plan but did not find a specific phase addressing this scenario. Perhaps this is intended to be handled in a later phase ? or could be treat in other issue

Thanks  @BraCR10  for testing. You're right that those issues exist, but they're both pre-existing behaviors not introduced by this PR:            
The storage scoping and the restore manager weren't changed here. They won't be covered in the remaining phases of this plan either, since they're about general account/session lifecycle.

 Would you mind opening separate issues for them? Otherwise I can do it. That way they get properly tracked and addressed in dedicated PRs. Thanks!

PR #529, opened 2026-03-13
Add P2P chat background notifications (Phase 2)

AndreaDiazCorreia:
## Summary
- Extract tradeKey event handling into `_handleTradeKeyEvent()` for clarity
- Add sharedKey matching in `_decryptAndProcessEvent()` so P2P chat messages arriving in background are detected
- Add `_handleP2PChatEvent()` that decrypts via `p2pUnwrap(sharedKey)` and triggers a generic "New message" notification
- Reuses existing localization keys and notification pipeline from Phase 1 (admin DM notifications)

## How it works
Previously, `_decryptAndProcessEvent()` only matched events by `tradeKey.public`, so P2P chat messages (which arrive at `sharedKey.public`) were silently dropped. Now, when no tradeKey match is found, sessions are checked by `sharedKey.public`. Matched events are decrypted with `p2pUnwrap` and produce the same generic chat notification used for admin DMs.

## Test plan
- [ ] Open an active trade with P2P chat between two users
- [ ] Send the app to background (do NOT kill it)
- [ ] From the counterparty, send a P2P chat message
- [ ] Verify a local notification appears with "New message" title and generic body
- [ ] Tap the notification and verify it navigates to the trade detail screen
- [ ] Verify Mostro protocol notifications (order updates, cancellations, timeouts) still work as before
- [ ] Verify admin/dispute chat background notifications still work as before
- [ ] Run `flutter analyze` — zero issues expected

## Summary by CodeRabbit

* **New Features**
  * Background notifications now support peer-to-peer chat messages and route to trade detail when applicable.

* **Improvements**
  * Notification processing tries multiple decryption paths for more reliable matching.
  * Distinguishes chat-message vs admin DM for correct title/message mapping.
  * Suppresses duplicate/self-notifications and logs missing order IDs safely.
  * Foreground app state prevents background notification retries while active.

* **Tests**
  * Added coverage for P2P chat notification routing and mapping.

### DIFF
--- a/lib/background/background.dart
+++ b/lib/background/background.dart
@@ -15,7 +15,15 @@ import 'package:mostro_mobile/services/logger_service.dart' as logger_service;
 import 'package:mostro_mobile/shared/providers/mostro_database_provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
-bool isAppForeground = true;
+/// Tracks whether the foreground app is currently active.
+///
+/// Defaults to `false` so that notifications fire by default when the
+/// background service is revived from a dead state (e.g. FCM wake after the
+/// app was killed), where no `app-foreground-status` message is sent.
+///
+/// Updated via the `app-foreground-status` message from the foreground app
+/// through `MobileBackgroundService.setForegroundStatus()`.
+bool isAppForeground = false;
 String currentLanguage = 'en';
 
 @pragma('vm:entry-point')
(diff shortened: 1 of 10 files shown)


Catrya (changes requested):
When my order went to Active status and i stared to receive DMs from my counterpert, I wasn't getting notifications for incoming messages until I closed and reopened the app. 

Then I started getting notifications for messages from the other person, but I was also getting a notification for every message I sent.

Then I closed the app again, and when I reopened it, I didn't get any message notifications from either the other person or me.

I closed and reopened it again, and I started getting notifications for incoming messages again, but also for all the messages I sent.

Catrya (commented):
After a couple of minutes with the app open, notifications stop coming in, even though I'm still receiving messages. I have to close and reopen it to get notifications for messages that arrive after that.

Even if I dismiss the notification at the top, I stop getting subsequent notifications.

Catrya (approved):
tACK

PR #533, opened 2026-03-16
fix: clarify timeout cancellation notification

### LINKED ISSUE #532: Improve cancellation message when order expires due to counterparty inactivity
Currently, when a user takes an order and the counterparty does not respond, the order expires after 15 minutes.  However, the message shown to the user says:
`"You cancelled the order"`

This is misleading because the user did not cancel the order. The order was automatically cancelled due to the counterparty not responding within the time limit. This can confuse users and make them think they performed an action they did not actually do. The message should clearly indicate that the order was cancelled because the counterparty did not respond.

**Suggested message**

Instead of:
`You cancelled the order`

Use something like:
`The counterparty did not respond in time. The order has been cancelled.`

**Expected behavior**
When an order expires due to inactivity from the counterparty, the message shown to the user should clearly reflect that the cancellation was caused by the timeout, not by the user.

Kemalau:
## What
- preserve timeout context for maker-side `canceled` events in notification data
- map timeout-driven cancellations to a dedicated localized notification message
- add unit coverage for the new cancellation context mapping

## Why
Fixes #532. When an order expires because the counterparty never responds, the current notification copy is too generic and can imply the user canceled the order themselves.

## Verification
- Added `test/features/notifications/notification_message_mapper_test.dart`
- Local Flutter verification could not be run in this environment because `flutter` is not installed on PATH here, so CI should run `flutter pub get`, code generation, and tests on the branch.

## Release Notes

* **New Features**
  * Orders canceled due to counterparty timeout now display a specific notification message, localized in English, Spanish, French, and Italian.

### DIFF
--- a/lib/features/notifications/utils/notification_data_extractor.dart
+++ b/lib/features/notifications/utils/notification_data_extractor.dart
@@ -148,8 +148,15 @@ class NotificationDataExtractor {
         break;
         
       case Action.canceled:
-        // Canceled orders don't generate persistent notifications
-        return null;
+        final order = event.getPayload<Order>();
+        final currentSession = session;
+
+        // Maker-side expirations arrive as canceled events for a pending order with no peer.
+        // Preserve that context so the notification copy can explain the timeout clearly.
+        if (order?.status == Status.pending && currentSession?.peer == null) {
+          values['cancellation_reason'] = 'peer_timeout';
+        }
+        break;
         
       case Action.cooperativeCancelInitiatedByYou:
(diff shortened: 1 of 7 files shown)


Catrya (changes requested):
The cancellation_reason: 'peer_timeout' is never set: regardless of which party timed out, the daemon always sends Action::Canceled with a None payload, so event.getPayload<Order>() returns null and the condition is never evaluated. getMessageKeyWithContext is also not updated to route to the new message key.

Issue #372, opened 2025-11-28
Error restoring chat: Null check operator used on a null value


Catrya:
If I restore a user who had orders with open chats, sometimes the chats don't appear, and if I click the contact button, I get this error: `Null check operator used on a null value`

<img width="720" height="761" alt="Image" src="https://github.com/user-attachments/assets/a1282967-8006-4f69-98bd-7d48c59054a6" />

If I have a new trade and open a chat for that order, the chat with the other user that wasn't working before now works.

To reproduce this:

- Have a previous user with at least one active order status, but with messages in the chat with the counterparty.

- Go to Account, create a new user, and without closing the app, import the previous user and go directly to the chat or My Trades and click Contact.

To avoid this: You must close the app and reopen it after creating a new user.
- You must close the app and reopen it after importing the user; only then will the chats appear.

These are the logs that appear: 
```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following _TypeError was thrown building ChatRoomScreen(dirty, dependencies:
[UncontrolledProviderScope], state: _ChatRoomScreenState#03ffa):
Null check operator used on a null value

The relevant error-causing widget was:
  ChatRoomScreen ChatRoomScreen:file:///home/catry/mobile/lib/core/app_routes.dart:146:28

When the exception was thrown, this was the stack:
#0      _ChatRoomScreenState.build (package:mostro_mobile/features/chat/screens/chat_room_screen.dart:51:31)
#1      StatefulElement.build (package:flutter/src/widgets/framework.dart:5833:27)
#2      ConsumerStatefulElement.build (package:flutter_riverpod/src/consumer.dart:539:20)
#3      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5723:15)
#4      StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5884:11)
#5      Element.rebuild (package:flutter/src/widgets/framework.dart:5435:7)
#6      ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:5705:5)
#7      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5875:11)
#8      ComponentElement.mount (package:flutter/src/widgets/framework.dart:5699:5)
...     Normal element mounting (157 frames)
#165    Element.inflateWidget (package:flutter/src/widgets/framework.dart:4548:16)
[...]

PR #538, opened 2026-03-17
fix: resolve chat history loss after reconnection or app restart

### LINKED ISSUE #536: [bug] Chat history partially lost after reconnection or app restart
Chat messages disappear partially after certain connectivity or lifecycle events. Two confirmed patterns:

1. After internet loss + reconnect: When the app lost internet mid-session and reconnected, only the most recent part of the chat was visible after refreshing. Older messages were gone.
2. After 24h without opening the app: On the following day, only the beginning of the chat was visible — recent messages had disappeared.

In both cases, using the shared key on Mostrix shows the complete chat, confirming the messages exist on the relay and the issue is client-side only.

To reproduce
Pattern 1:
1. Open an active trade chat
2. Lose internet mid-session (airplane mode or disconnect)
3. Reconnect and refresh the chat
4. Expected: full chat history. Actual: only tail of conversation visible

Pattern 2:
1. Have an active trade chat with messages
2. Close the app for ~24h
3. Reopen the app and navigate to the chat
4. Expected: full history. Actual: only the beginning of the conversation visible

**Additional context**
The shared key works correctly in Mostrix, confirming the data is on the relay. The bug is in how the app loads or preserves messages locally.

Catrya:
fix #536 
  - `reload()` now calls `_loadHistoricalMessages()` before re-subscribing, so messages stored while the app was in background are loaded into state on foreground resume
  - `sendMessage()` persists the wrapped event to disk immediately after successful publish, preventing message loss  if the relay echo never arrives                                                                                     
  - `reloadAllChats()` awaits all parallel `reload()` completions before calling `refreshChatList()`, eliminating the race condition where chats appear empty and get filtered out 
 

Test:                                                                                                    
                                                            
  - Open a trade chat with several messages, go to airplane mode, reconnect — verify full chat history is visible 
  - Send a message and immediately kill the app before the relay echo arrives — reopen and verify the sent message persists                                                                                                           
  -Background the app for several minutes while messages arrive, resume — verify all messages appear without missing any                                                                                                         
  - Restore a user with active orders — verify chats load correctly without null errors (no regression on #372)

## Summary by CodeRabbit

* **Bug Fixes**
  * Persist outgoing messages to local storage immediately after send, then update the UI to avoid lost optimistic updates.
  * Load historical messages before (re)subscribing so returning to the app shows complete conversations and avoids empty chats.
  * Gate incoming/outgoing events by ownership earlier to prevent invalid or duplicate writes.
  * Reload all chats in parallel and await completion (including during app foregrounding) to prevent race conditions.

* **Documentation**
  * Added architecture docs describing the P2P chat design, storage, and message flows.

### DIFF
--- a/docs/architecture/P2P_CHAT_SYSTEM.md
+++ b/docs/architecture/P2P_CHAT_SYSTEM.md
@@ -22,7 +22,7 @@ For the **protocol specification** (NIP-59, ECDH, event format), see the [Mostro
 
 ## Message Flow: Receiving
 
-```
+```text
 Relay
   │  kind 1059 gift wrap events (encrypted)
   ▼
@@ -64,7 +64,7 @@ The relay sends events for all chats through this single subscription. Events ar
 
 ## Message Flow: Sending
 
-```
+```text
 User types message
   │
   ▼
@@ -161,7 +161,7 @@ On app restart:
 
 ### App startup (`app_init_provider.dart`)
 
-```
+```text
 1. NostrService.init()         → relay connections
 2. KeyManager.init()           → crypto keys from secure storage
 3. MostroNodes.init()          → node metadata
@@ -199,7 +199,7 @@ When the app returns to foreground after losing connection:
 
 ## Historical Loading (`_loadHistoricalMessages`)
 
-```
+```text
 Sembast query: WHERE type = 'chat' AND order_id = orderId
   │
   ▼


BraCR10 (commented):
Hi @Catrya,

While testing the PR, I found an issue, could you check it?:

1. I started a chat with some messages.
2. I turned off the Wi-Fi on my phone.
3.  I waited a few seconds and turned it back on.
4.  Even though the app was able to send messages and the other phone received them, the phone that lost the connection no longer received messages.

After restoring the orders, the chat worked properly again.

BraCR10 (approved):
LGTM

PR #571, opened 2026-04-09
feat: add configurable trade history retention in settings

Catrya:
- Add session expiration selector with 6 presets: 1 week, 1 month, 3 months, 6 months, 1 year, never
  - Default remains 1 month (720 hours), stored in Settings and persisted via SharedPreferences
  - SessionNotifier reads expiration from user settings with Config fallback
  - Value 0 means never delete, protected by explicit guards in init() and cleanup()
  - New Trade History card in settings screen between Notifications and Dev Tools
  - Localization keys added for EN, ES, IT, FR, DE
  - 11 unit tests covering serialization, presets, backward compatibility, and never mode
  - Includes log output when user changes the setting

## Summary by CodeRabbit

* **New Features**
  * User-configurable session expiration (hours) with presets (1w/1m/3m/6m/1y/never); new Trade History settings card and selector; setting is persisted.

* **Behavior**
  * Expiration now cascades deletion of expired sessions and associated data; “never” loads all sessions (no cutoff).

* **Internationalization**
  * Added DE/EN/ES/FR/IT translations for trade-history UI and options.

* **Tests**
  * Added tests for defaults, updates, clearing, serialization, and presets.

* **Documentation**
  * Updated docs covering retention, “never” mode, and cleanup.

### DIFF
--- a/docs/architecture/APP_INITIALIZATION_ANALYSIS.md
+++ b/docs/architecture/APP_INITIALIZATION_ANALYSIS.md
@@ -286,7 +286,7 @@ Future<void> init() async {
 
 **Critical Aspects**:
 - Loads sessions from Sembast database
-- Filters expired sessions (older than 720 hours / 30 days)
+- Filters expired sessions based on user-configured retention period (default: 720 hours / 30 days, configurable in Settings). Expired sessions and all associated data (events, messages, notifications, read status) are cascade-deleted. If set to "never", all sessions are loaded without filtering.
 - Updates `state` which triggers all listeners
 - **Must complete before SubscriptionManager setup**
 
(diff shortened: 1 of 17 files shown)


Catrya:
@coderabbitai review

AndreaDiazCorreia (changes requested):
@Catrya, I've been testing this branch and wanted to share a few things.

One thing that happened to me: when I started the app on this branch, it deleted my previous session right away, and I don't think that should happen.

Another thing, it seems like it's not checking whether there are active trades, and I think it would be a good idea, just as a precaution, to evaluate that before deleting anything.

And one more, about UX: the text that's currently in the info message of the widget should actually be shown as a confirmation dialog before applying the change, so the user confirms they understand what this change means.

Catrya:
@coderabbitai review

AndreaDiazCorreia (approved):
tACK

grunch (approved):
tACK

PR #588, opened 2026-05-06
feat(restore): differentiate restore outcomes by cant-do reason

Catrya:
Map cant-do responses from Action.lastTradeIndex to specific user-facing messages instead of a single generic error:

- not_found: complete restore as new user with informative subtext ("No previous data for this user in this Mostro"). Same green icon as a normal successful restore.
- invalid_trade_index: show the error overlay with a specific subtext ("Invalid trade index").
- Any other cant-do reason or timeout: keep the existing generic error message.

Restore behavior (key indices, sessions, navigation) is unchanged. syncTradeIndex remains silent.

## Summary by CodeRabbit

* **New Features**
  * Restore flow now detects when no prior transaction history exists and completes as a new user, showing a tailored status line and briefly extending the message display.
  * Restore errors for an invalid trade index now surface a specific localized error line.

* **Documentation**
  * Added localized restore messages for "no previous data" and "invalid trade index" in English, German, Spanish, French, and Italian.


Catrya:
@coderabbitai review

Catrya:
@coderabbitai review

BraCR10 (approved):
tACK

While I was testing, I noticed an unusual behavior. However, I think it is outside the scope of this PR.

First, I created an order in one Mostro instance. Then, I connected to another node where my user did not have any orders. Finally, I tried to reconnect to the original Mostro instance where my orders existed.

I tried more than three times and it did not work. However, after each failed restore attempt while reconnecting to the instance, I manually refreshed the user, and the restore worked properly afterwards.

grunch (approved):
LGTM

PR #602, opened 2026-05-22
feat(notifications): persist order cancellations in history with context

### LINKED ISSUE #535: Add notification screen when orders are cancelled due to counterparty inactivity
When an order is cancelled due to counterparty inactivity while in states such as: 
`waiting_buyer_invoice`,  `waiting_payment`
The app currently shows a SnackBar message informing the user that the order was cancelled. However, this event does not appear in the notifications screen (the screen accessed through the bell icon). Because of this, the user briefly sees the SnackBar but cannot later review what happened to the order.

If a user is the taker, completes their step, but the maker does not continue, the order is eventually cancelled.
The taker should receive a notification in the notifications screen indicating that the order was cancelled.

**Example message:**
The seller did not continue the order. The order has been cancelled.

AndreaDiazCorreia:
## Summary

Closes #535.

When an order is canceled due to counterparty inactivity in `waiting-payment` or `waiting-buyer-invoice`, the user previously only saw a transient SnackBar — the event never reached the notifications screen (bell icon), so they could not review what happened to the order afterwards.

This PR persists every `Action.canceled` event to the notification history with a context-aware message based on the order's previous status:

- `waiting-payment` → "The seller did not continue the order. The order has been cancelled."
- `waiting-buyer-invoice` → "The buyer did not continue the order. The order has been cancelled."
- Other (manual cancel, pending expiry, etc.) → generic "The order has been canceled."

## Changes

- **`NotificationDataExtractor`**: new `previousStatus` parameter; `Action.canceled` no longer returns `null` (which previously skipped persistence) and instead emits `NotificationData` with `previous_status` in `values`.
- **`NotificationMessageMapper.getMessageKeyWithContext`**: resolves the correct localization key for `Action.canceled` based on `previous_status`.
- **`AbstractMostroNotifier.subscribe()`**: captures `previousStatus = state.status` before `state.updateWith(msg)` and forwards it through `handleEvent` → extractor.
- **`AbstractMostroNotifier.handleEvent()`**: removed the now-redundant `showCustomMessage('orderCanceled')` call — the centralized `notify()` path already shows the SnackBar **and** persists to history.
- **`OrderNotifier`**: forwards `previousStatus` in its `handleEvent` override and converts the pending-order-expiry `showCustomMessage('orderCanceled')` to a `notify(Action.canceled, ...)` call so it also appears in the history.
- **Localizations**: added `notification_order_canceled_by_seller_inactivity_message` and `notification_order_canceled_by_buyer_inactivity_message` to `en`, `es`, `it`, `de`, `fr`.

## Test plan

- [ ] `fvm flutter analyze` → no issues (verified)
- [ ] `fvm flutter test` → all tests pass (verified, 369/369)
- [ ] Manual: take a sell order, let the seller (maker) not pay → after `Action.canceled`, open the notifications screen and verify the "seller did not continue" message appears
- [ ] Manual: take a buy order, let the buyer (maker) not provide an invoice → verify the "buyer did not continue" message appears
- [ ] Manual: create a pending order and let it expire → verify a generic cancellation entry appears in the notifications screen
- [ ] Manual: verify a single SnackBar is shown (no duplicate) on cancellation
- [ ] Manual: verify message is correctly localized in EN, ES, IT

## Summary by CodeRabbit

* **New Features**
  * Cancellation notifications are always produced and persisted; when available they include the previous order status and correctly distinguish user-initiated vs automatic cancels.
  * Notifications now use distinct messages for cancellations due to seller vs buyer inactivity.
  * Manual cancels are tracked/untracked so user-initiated and automatic expirations are handled distinctly; automatic expirations generate deterministic notification events.

* **Localization**
  * Added localized messages for inactivity-based cancellations in English, German, Spanish, French, and Italian.

### DIFF
--- a/lib/features/notifications/utils/notification_data_extractor.dart
+++ b/lib/features/notifications/utils/notification_data_extractor.dart
@@ -10,7 +10,7 @@ import 'package:mostro_mobile/shared/providers.dart';
 class NotificationDataExtractor {
   /// Extract notification data from MostroMessage
   /// If ref is null, will use fallback methods for nickname resolution
-  static Future<NotificationData?> extractFromMostroMessage(MostroMessage event, Ref? ref, {Session? session, Status? previousStatus}) async {
+  static Future<NotificationData?> extractFromMostroMessage(MostroMessage event, Ref? ref, {Session? session, Status? previousStatus, bool wasUserInitiatedCancel = false}) async {
     Map<String, dynamic> values = {};
     bool isTemporary = false;
     
@@ -148,10 +148,12 @@ class NotificationDataExtractor {
         break;
         
       case Action.canceled:
-        // Persist cancellations in notification history with a variant message
-        // based on the order's previous status, so the user can review what
-        // happened to the order later from the notifications screen.
-        if (previousStatus != null) {
+        // Persist cancellations in notification history. Only attach
+        // previous_status (which drives the inactivity-specific message)
+        // when the cancel was NOT user-initiated; manual cancels in
+        // waiting-payment / waiting-buyer-invoice should fall back to the
+        // generic message instead of falsely blaming the counterparty.
+        if (!wasUserInitiatedCancel && previousStatus != null) {
           values['previous_status'] = previousStatus.value;
         }
         break;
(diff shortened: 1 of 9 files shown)


Catrya (approved):
tACK

PR #604, opened 2026-05-26
feat(bond): show the taker when their anti-abuse bond is slashed. Phase 4 

Catrya:
Handles the `bond-slashed` notice that Mostro sends to a taker whose anti-abuse bond is slashed on a waiting-state timeout, and surfaces it to the user.

  - Recognize the `bond-slashed` action so the message no longer fails to parse and gets dropped; its SmallOrder (bond
   amount, null status) never overwrites the tracked order/status.
  - Persist a notification and show it in the notifications screen; tapping it opens a dialog with the order details
  in prose and a green close button.
  - Defer deleting a canceled bonded order's session for 60s so the trailing bond-slashed notice can still be received
   and decrypted; release the session as soon as it arrives. Out-of-order safe.
  - Add OrderState regression tests for bond-slashed.

## Summary by CodeRabbit

* **New Features**
  * Added bond-slashed notifications with dedicated dialog and detailed fields (amount, order ID, payment method, fiat info).
  * Notification details view omits empty detail blocks for cleaner display.

* **Bug Fixes / Behavior**
  * Preserve order status when bond-slashed notices arrive.
  * Deferred deletion of canceled bonded sessions with restart reconciliation; user-initiated cancels delete immediately.

* **Localization**
  * Added bond-slashed strings in English, German, Spanish, French, and Italian.

* **Tests**
  * Added tests for bond-slashed order handling and notification tap/dialog behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MostroP2P/mobile/pull/604?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Catrya:
> I re-reviewed the current head. The restart/reconcile work is a good improvement, but I do see one remaining correctness issue that I would fix before merge.
> 
> Blocking issue:
> 
> * `OrderNotifier.cancelOrder()` marks the order as user-initiated **after** awaiting `mostroService.cancelOrder(orderId)`. That creates a race with the incoming `Action.canceled` event: if the daemon responds quickly and the event is processed before that `await` returns, the `canceled` handler will not see the user-initiated flag yet. In that case it can wrongly take the timeout-slash path and defer session deletion for 60 seconds on a voluntary cancel, which is exactly the case this patch is trying to avoid.
> 
> I think the intent here is right, but the marker needs to be set before the async send (and then cleared/reverted if the send itself fails) so the event path can observe it reliably.
> 
> The rest of the new direction looks reasonable to me: extracting the decision logic into pure helpers was a good move, the retake cleanup guard makes sense, and the added tests improve confidence. But I would still block on the race above because it changes cancel behavior in a real timing-dependent way.

I don't think the race is reachable, and moving the marker before the send would actually reintroduce the bug this phase fixes.

  There are two sources of Action::Canceled to the taker (confirmed in the daemon):
  1. Voluntary cancel — releases the bond, republishes, no bond-slashed.
  2. Waiting-timeout slash — daemon autonomously sends Canceled, then bond-slashed ~150ms later.

For a real voluntary cancel, the Canceled is the daemon's response to the event we publish, so it's causally after publishEvent. markUserInitiatedCancel runs in the microtask continuation right after that await (no await between), and incoming relay events are I/O macrotasks that can't preempt the microtask drain, so the flag is always set first. The "voluntary cancel wrongly deferred" path isn't reachable.

  The only window where the flag isn't set yet is when an autonomous timeout-slash Canceled overlaps a user cancel, and there, deferring is correct, because a bond-slashed is coming:

  - Marker after send (current): userInitiated=false → defer → bond-slashed arrives → notice surfaced. ✅
  - Marker before send (proposed): userInitiated=true → delete immediately → trade key dropped → trailing bond-slashed
  can't be decrypted → notice lost. ❌

  revert on failure doesn't help, since the cancel send succeeds — the daemon just slashed first. The current ordering is the correct one; I can add a comment documenting why the marker is set after the send.

PR #608, opened 2026-06-03
feat(bond): maker anti-abuse bond on order creation (Phase 5)

Catrya:
Implements the mobile side of anti-abuse bond Phase 5: when a node has the bond enabled for makers (`apply_to = make | both`), creating an order now requires paying a bond before the order is published.

## What this does
  - `AddOrderNotifier` now handles `pay-bond-invoice` during creation: cancels the create timeout, shows the pay-bond
  screen, and stays alive so the post-bond `new-order` ack still confirms via `_confirmOrder`.
  - The uncommitted order stays ephemeral: its session lives in memory only (new `registerSessionInMemory`) and a
  transient, non-serialized `Session.bondPending` flag keeps the shared pay-bond handler from persisting it. An
  abandoned order leaves nothing on disk and never reappears after a restart; it is also never shown in My Trades (no
  published NIP-33 event).
  - Cancel for the maker abandons locally (clears the stored bond message + in-memory session) instead of sending a
  `cancel` the daemon rejects during `WaitingMakerBond`; the bond hold invoice expires server-side. Taker bonds and
  published orders keep the existing cancel flow.
  - Maker-specific bond screen copy warning the user to keep the screen open and pay, or the order won't be created.
  ## How to test manually

  Requires a Mostro node with the anti-abuse bond enabled for makers (`[anti_abuse_bond] enabled = true`, `apply_to =
  make` or `both`). The info event should show `bond_enabled: true`.

  ### Bond-enabled node (maker)
  1. **Happy path** — Create a sell order. The pay-bond screen appears with a QR and the maker-specific copy ("Before  your order can be created…"). Pay the bond invoice from a Lightning wallet → the order is published and you land on the order-created confirmation. The order shows in My Trades as pending.
  2. **No false timeout** — Confirm the old "no response received" message no longer appears while the bond invoice is outstanding (it used to fire after 10s).
  3. **Cancel button** — Create another order; on the pay-bond screen tap Cancel and confirm. You return to the order book. The order is not created, nothing lingers in My Trades, and no `cancel` is sent to Mostro (the bond invoice just expires server-side ~5 min later).
  4. **Leave the screen** — Create an order, reach the pay-bond screen, then press back without paying. The order is
  not in My Trades.
  5. **Ephemeral on restart** — Create an order, reach the pay-bond screen, do NOT pay, kill the app and reopen.
  Nothing is waiting and nothing shows in My Trades; you must create the order again.

  ### Bond-disabled node (regression)
  6. Point the app at a node without the bond (or `apply_to = take`). Create an order → it publishes immediately with
  no pay-bond screen, exactly as before.

  ### Taker (regression)
  7. On the bond-enabled node, take an existing order. The taker pay-bond flow is unchanged: the bond screen shows the taker copy, paying proceeds, and tapping Cancel sends a cancel to Mostro and frees the order.

  Nodes without the bond are unaffected: the `new-order` path is untouched, so creation behaves exactly as before.

## Summary by CodeRabbit

* **New Features**
  * Maker bond payment flow: transient in-memory sessions used during bond payment and a dedicated pay-bond screen flow with maker-specific messaging and navigation.

* **Bug Fixes**
  * Session persistence and cancellation now respect transient bond state so unfinished bond payments aren’t persisted or treated as final.

* **Localization**
  * Updated/added bond-related strings in EN/DE/ES/FR/IT, including maker-specific guidance.

* **Tests**
  * Added tests for transient bond flag, maker bond state transitions, and timeout cleanup behavior.


Catrya:
Added a regression test (maker_bond_timeout_test.dart) for this. I don't think the issue is reachable: the create-order timeout is cancelled on the first line of _handleMakerBondInvoice, the same method that sets bondPending and navigates to the pay-bond screen. Since the abandon branch in cancelOrder() is only reachable from that screen, the timer is always already disarmed before abandon runs — the two states can't coexist. Also, OrderNotifier doesn't hold the requestId (it's keyed by orderId), so cancelling that timer from the abandon branch isn't even possible, nor needed.

The test confirms the mechanism it relies on: cancelSessionTimeoutCleanupForRequestId disarms the timer and a second cancel is a safe no-op.

grunch (approved):
utACK

PR #610, opened 2026-06-09
feat(relays): bootstrap relay discovery instead of pinned default relays

### LINKED ISSUE #609: Add hidden bootstrap relays used only to discover Mostro's kind 10002, never shown to the user
PR #607 added a second hardcoded default relay (wss://mostro-p2p.tech). These hardcoded relays are loaded as RelaySource.defaultConfig: persisted, always present in the user's relay list, always connected, and surviving sync unless blacklisted. This causes idle connections to relays the active Mostro may not use, divergence between the visible list and Mostro's endorsed relays, and UX special-casing in relay_selector.dart.
This issue replaces that approach with a defensive, internal-only bootstrap mechanism.

###  What will be done

Introduce a hidden, internal list of bootstrap relays that the app uses behind the scenes, only when needed, with the sole purpose of fetching the Mostro instance's kind 10002 relay list. Once the real relays are discovered from the 10002, those become the relays loaded in the app and shown to the user, and all subsequent queries go exclusively through them, not through the bootstrap relays.

### When bootstrap is activated

  Bootstrap must not run on every launch. The relays discovered from the last kind 10002 are already persisted (settings.relays, RelaySource.mostro), so a normal restart simply loads and uses them. The central condition is: do we currently have at least one discovered relay that connects? If yes, bootstrap never runs.

  On startup:
  1. Load the persisted discovered relays (from the last 10002).
  2. Attempt to connect to them.
  3. If ≥1 connects within a 5–8s timeout → normal operation, bootstrap is not used at all.
  4. If 0 connect (all unreachable) within the 5–8s timeout → activate bootstrap to re-fetch the 10002.

  First launch ever (no persisted relays): activate bootstrap immediately, since there is nothing to load.

  During an active session: if all discovered relays drop and reconnection fails, re-activate bootstrap automatically to recover connectivity and re-fetch the 10002. Bootstrap is evaluated both at startup and at runtime, but only as a total-failure fallback.

  Normal 10002 updates are unaffected: when Mostro updates its relay list, the new 10002 arrives in real time through the already-connected discovered relays (existing real-time sync). Bootstrap plays no part in normal updates, it is strictly the cold-start / total-failure path.

  Discovery and retirement

  1. Discovery phase: When activated, the app connects to the bootstrap relays solely to fetch Mostro's kind 10002 event.
  2. Load discovered relays: The relays found in the 10002 are loaded into the app and are the only ones shown to the user.
  3. Operate through discovered relays: After discovery, all event fetching/publishing goes through the discovered relays, not the bootstrap relays.
  4. Retirement: Bootstrap connections are dropped once a valid kind 10002 has been received and at least one discovered relay is confirmed reachable.
  5. Mostro offline / no 10002: If the 10002 never arrives and no discovered relay is alive, the app keeps retrying the bootstrap relays with a timeout/backoff, staying active until something responds, so the app is never left without a path to discover relays.

  Constraints

  - Bootstrap relays are fully hidden: never shown in the UI or settings, never persisted into the user's relay list. They are pure internal infrastructure.
  - Bootstrap connections live outside RelaysNotifier state, so they never trigger the promotion/elimination sync logic and never appear among user/Mostro relays.
  - The bootstrap list is a dedicated internal constant, separate from the relays persisted for a given instance.

  Cleanup enabled by this change:

  - Removes the always-present idle defaultConfig connections introduced as defaults.
  - Removes the wss://relay.mostro.network special-casing in relay_selector.dart (~line 250), since bootstrap relays never reach the UI.

  Scope:  Architectural change touching NostrService + RelaysNotifier. Intentionally separate from PR #607, which ships the hardcoded defaults as an interim mitigation.

Catrya:
fix #609
 Replaces the hardcoded default relay that was always seeded into the user's relay list with a hidden bootstrap discovery layer. The visible relay list now contains only relays discovered from the Mostro's kind 10002 plus user-added relays. A small set of defensive bootstrap relays is used only to discover that list: at cold start, when every discovered relay is unreachable, or when switching Mostro instances.

  ## Why

  The default relay was persisted and always connected even when the active Mostro did not use it, leaving idle connections, diverging the visible list from what the Mostro endorses, and requiring UI special-casing.

  ## How

  - `Config.bootstrapRelays` (relay.mostro.network, mostro-p2p.tech, nos.lol, relay.damus.io) live only at the NostrService connection level: never persisted, never shown.
  - Cold start: when settings has no relays, NostrService.init falls back to the bootstrap relays to fetch the kind 10002.
  - All-down recovery: RelayHealthMonitor (6s watchdog) re-engages bootstrap and re-subscribes (subscribeAll + kind 10002) when no relay is alive.
  - Instance switch: clears the list and ensures bootstrap connectivity for the new instance's kind 10002.

  All engagement is additive (never disconnectFromRelays), so existing connections and subscriptions are never dropped. Retirement is logical only since dart_nostr has no per-relay disconnect.

  ## How to test manually

  Watch logs with `flutter logs` (or `adb logcat | grep -i relay`).

  1. Cold start: clear app data and launch. Logs show "falling back to bootstrap relays", then "Discovered N relays from Mostro 10002". Settings > Relays lists only the discovered relays (not nos.lol / relay.damus.io). The order book loads.
  2. Normal restart: close and reopen. The relay list appears immediately and logs do not show a bootstrap connection (discovered relays work).
  3. Instance switch: change the Mostro node. Logs show bootstrap connection then discovery of the new instance's relays; the list updates.
  4. Manual relay: add a valid wss relay. It persists and is not removed by a later kind 10002 sync.
  5. No event loss: create or take an order and exchange chat messages; events arrive normally. 
  6. All-down recovery (optional): toggle WiFi off and on. Logs show "No live relays detected; engaging bootstrap relays" and events keep flowing after the network returns.

## Summary by CodeRabbit

* **New Features**
  * Added a relay health watchdog that auto-recovers connectivity and re-subscribes when relays are down.
  * Added defensive bootstrap discovery to find and activate relay lists when no user relays are available.

* **Improvements**
  * Relay initialization now tolerates empty configured relay lists and uses bootstrap discovery as needed.
  * Default settings start with no pre-seeded relays.

* **Bug Fixes**
  * Streamlined relay deactivation flow in the UI (removed special-case confirmation for default relays).

* **Documentation**
  * Added architecture docs describing the bootstrap discovery layer and safety invariants.

* **Tests**
  * New coverage for relay health recovery, bootstrap behavior, and relay connection invariants.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,6 +74,7 @@ When implementing or debugging protocol-related features (order flows, actions, 
 - **Source Tracking**: Relays tagged by source (user, mostro, default) for appropriate handling and storage strategy
 - **Smart Re-enablement**: Manual relay addition automatically removes from blacklist, Mostro relay re-activation removes from blacklist during sync
 - **URL Normalization**: All relay URLs normalized by removing trailing slashes to ensure consistent matching
+- **Bootstrap Discovery Layer**: Hardcoded relays are NOT seeded into the visible list; `Config.bootstrapRelays` are defensive connections used only to discover a Mostro's kind 10002 (cold start via `NostrService.init` fail-safe, all-down via `RelayHealthMonitor` watchdog, instance switch via `ensureBootstrapConnectivity()`). Engaged additively (never `disconnectFromRelays()`); retirement is logical only since dart_nostr has no per-relay disconnect. See `docs/architecture/RELAY_SYNC_IMPLEMENTATION.md` ("Bootstrap Relay Discovery Layer")
 - **Implementation**: Located in `features/relays/` with core logic in `RelaysNotifier`
 
 #### Manual Relay Addition
--- a/docs/architecture/RELAY_SYNC_IMPLEMENTATION.md
+++ b/docs/architecture/RELAY_SYNC_IMPLEMENTATION.md
@@ -45,6 +45,37 @@ The system handles NIP-65 relay list events (kind 10002), implements dual storag
 
 ---
 
+## Bootstrap Relay Discovery Layer
+
+The app does not seed hardcoded "default" relays into the user-visible relay list. Instead, a small set of **defensive bootstrap relays** (`Config.bootstrapRelays`) is used only to discover a Mostro instance's kind 10002 relay list. Once discovered, the app operates exclusively over the discovered relays; the bootstrap relays are never persisted nor shown.
+
+### Design
+
+- **Bootstrap relays live only at the NostrService connection level**, never in `RelaysNotifier` state or `settings.relays`. The visible list and `settings.relays` always contain only discovered (Mostro) relays plus user relays. This keeps the user-facing relay list equal to what the Mostro instance actually endorses, instead of pinning hardcoded relays the instance may not use.
+- **Engaged only when needed**, via additive init (never disconnect):
+  - **Cold start**: when `settings.relays` is empty, `NostrService.init()` falls back to `Config.bootstrapRelays` so the kind 10002 can be fetched.
+  - **All-down**: `RelayHealthMonitor` (a periodic watchdog, `Config.relayDiscoveryTimeout` interval) detects `liveRelayCount == 0` and calls `ensureBootstrapConnectivity()` + `subscribeAll()` + re-subscribes the kind 10002.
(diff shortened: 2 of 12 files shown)


grunch (approved):
great job! utACK

PR #621, opened 2026-06-18
feat: add dual receive for transport v2 (Phase A)

AndreaDiazCorreia:
> [!IMPORTANT]
> Depends on #620 (transport v2 migration spec). That PR must be merged first.

Implements **Phase A — Dual receive** of the transport v1 → v2 migration described in `docs/architecture/TRANSPORT_V2_MIGRATION.md`.

Adds the machinery to receive and decrypt protocol-v2 NIP-44 direct messages (kind 14) alongside the existing v1 gift-wrap path (kind 1059), selecting per node.

## Changes
- **`Transport` enum + resolver** (`lib/features/mostro/transport.dart`): single source of truth for the per-node transport. Phase A always resolves to `giftWrap`; Phase C wires real `protocol_version` resolution.
- **Transport-aware order filter** (`subscription_manager.dart`): v2 nodes subscribe to kind 14 pinned to `authors=[mostroPubkey]` and `p=[tradeKeys]`; v1 keeps kind 1059.
- **v2 unwrap** (`nostr_utils.dart`): `decryptNIP44DirectEvent` verifies the kind-14 author signature (author = node), NIP-44 decrypts the content, and returns the message tuple.
- **Receive branch** (`mostro_service.dart`): `_onData` dispatches on `event.kind`; v1 and v2 converge on `MostroMessage.fromJson(tuple[0])`.

## Scope
Dormant by design: the resolver returns `giftWrap`, so the emitted filters and runtime behaviour are identical to today. The v1 gift-wrap path is byte-for-byte unchanged. Phase C activates dual-receive via auto-detection.

Out of scope: dual send (Phase B), `protocol_version` detection/wiring (Phase C), formal tests (Phase D), NIP-17 peer/dispute chat.

## Verification
- `flutter test`: matches baseline (pre-existing unrelated failures only); v1 path unchanged.

## Release Notes

* **New Features**
  * Added support for Mostro protocol v2 alongside v1 for wire transport, with automatic per-node detection.
  * Updated orders and trade-key notifications to use protocol-appropriate handling (including v2 direct replies).
  * Improved DM processing for protocol-v2 messages during the migration window.

* **Documentation**
  * Refreshed the transport migration architecture guide with clarified phase coverage and behavior.

* **Tests**
  * Added unit tests for protocol version parsing and transport resolution.

### DIFF
--- a/docs/architecture/TRANSPORT_V2_MIGRATION.md
+++ b/docs/architecture/TRANSPORT_V2_MIGRATION.md
@@ -251,6 +251,7 @@ for either transport without the two drifting apart.
 | Publish + PoW + recipient | `lib/services/mostro_service.dart:338-360` |
 | Receive subscription filters (kind 1059) | `lib/features/subscriptions/subscription_manager.dart:121-160` |
 | Receive decrypt (unWrap → `result[0]`) | `lib/services/mostro_service.dart:129-155` |
+| Background receive decrypt (unWrap → `result[0]`) | `lib/features/notifications/services/background_notification_service.dart:199-322` |
 | NIP-59 unwrap | `lib/shared/utils/nostr_utils.dart:383-443` |
 | Node info / `protocol_version` parse | `lib/features/mostro/mostro_instance.dart:119-251` |
 | Global `version` constant | `lib/core/config.dart:56` |
@@ -276,6 +277,14 @@ unchanged.
 - Branch `MostroService` receive (`lib/services/mostro_service.dart:129-155`):
   v1 yields an inner rumor whose content is the 2-tuple; v2's decrypted content
   **is** the tuple directly. Both converge on `MostroMessage.fromJson(tuple[0])`.
+- Apply the **same receive branch to the background isolate**
+  (`lib/features/notifications/services/background_notification_service.dart`):
+  accept kind `14` in `_decryptAndProcessEvent` and branch `_handleTradeKeyEvent`
+  on `event.kind`. The background isolate has no Riverpod settings provider, so
+  the node pubkey (the v2 author to verify) is read from persisted settings
+  (`SharedPreferencesKeys.appSettings`). Without this, v2 replies received while
+  the app is backgrounded would be silently dropped once Phase C flips the
+  transport.
 
 ### Phase B — Dual send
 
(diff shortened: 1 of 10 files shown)


AndreaDiazCorreia:
@codex review

grunch (dismissed):
## Review — Phase A dual receive

Reviewed the full diff at head `954a7bd1` plus every review thread. Walked the foreground receive path, the background isolate, the subscription/transport resolution, the NIP‑44 decrypt helper and the new tests. `flutter analyze` is clean on all 7 changed `lib/` files and the two new test files pass (27/27).

### Earlier review threads — all addressed
- **Background kind‑14 path** (codex / ermeme): fixed. `_decryptAndProcessEvent` now allows kind 14, `_handleTradeKeyEvent` branches to `decryptNIP44DirectEvent`, and `_loadMostroPubkey` reads the node pubkey from persisted `AppSettings` inside the isolate (no Riverpod there). The orders filter is transferred to the background service via `lifecycle_manager`, so a v2 node yields a kind‑14 filter there too.
- **`resolveTransport(null)` hardcoded** (ermeme): fixed. `_resolveOrdersTransport()` now reads the real `protocol_version` from the kind‑38385 info event, and `_initMostroInstanceListener` re‑subscribes when the info event arrives after the initial subscription. Covered by `transport_test.dart` and the `protocol_version` cases in `mostro_instance_test.dart`.
- **Doc status line** (CodeRabbit): fixed — now states Phase A is implemented.

### On the remaining "storage key is transport‑dependent" blocker
I dug into this one and I don't think it holds up — and the suggested remedy would actually break history.

The persisted record key (`decryptedId ?? event.id`) must be **unique per message**, because a single order has many messages. Every history query keys on the logical `id` **field** of the stored `MostroMessage` (`Filter.equals('id', orderId)` in `mostro_storage.dart`), which is the order id and is fully transport‑invariant. The record key is only a per‑message uniqueness/dedup handle.

So persisting "under the Mostro message/order identity from `tuple[0]`" would set the record key to the order id, and `addMessage`'s `if (await hasItem(id)) return;` would then store **only the first message per order** and silently drop the rest. The current scheme (inner rumor id for v1, outer kind‑14 event id for v2) is the correct design: distinct per‑message keys, with logical grouping done by the `id` field at query time.

Cross‑transport divergence also can't occur in practice: a node speaks exactly one transport, and within a transport the identifier is stable across relays, so dedup is sound. History identity is already transport‑invariant where it matters (the `id` field).

### Other things I checked and found fine
- v2 filter pins `authors:[mostroPublicKey]` + `p:[tradeKeys]`, disambiguating from NIP‑17 peer chat (also kind 14).
- `decryptNIP44DirectEvent` verifies kind, expected author, and recomputes the event id + Schnorr signature before decrypting — author signature is load‑bearing for v2, good.
- `event.recipient` reads the `p` tag, so session matching works identically for kind 14.

LGTM. Approving.

PR #622, opened 2026-06-18
feat(nwc): allow paying invoice from another wallet on insufficient balance

### LINKED ISSUE #476: nwc: Allow payment with another wallet
When the order is in "waiting-payment" status, there should be a "Show Invoice" button. This way, if the wallet connected to NWC doesn't have enough funds or couldn't pay, the user can pay the invoice from another wallet and the order won't be canceled.

Image

AndreaDiazCorreia:
## Summary

When paying via NWC, the user is no longer stuck if the connected wallet has insufficient funds or the payment fails.

- Always show a "Show invoice" button in the NWC payment flow so the user can pay the invoice from another wallet, keeping the order alive instead of having to cancel.
- Add a manual refresh button next to the wallet balance so it updates instantly after topping up the wallet, without waiting for the periodic refresh.
- Add `nwcShowInvoice` localization key (en/es/it/fr/de).

Closes #476

### DIFF
--- a/lib/l10n/intl_de.arb
+++ b/lib/l10n/intl_de.arb
@@ -1538,6 +1538,7 @@
     "nwcQuotaExceeded": "Ausgabenlimit der Wallet überschritten",
     "nwcRetryPayment": "Zahlung wiederholen",
     "nwcPayManually": "Stattdessen manuell zahlen",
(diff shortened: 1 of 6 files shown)


Catrya (approved):
tACK

PR #623, opened 2026-06-18
fix(restore): show new-invoice prompt for failed payout in settled-hold-invoice

### LINKED ISSUE #615: Restore shows "released / paying sats" instead of the new-invoice prompt when the LN payout failed
### Summary

  When Mostro's Lightning payment to the buyer **fails**, the order stays at status `settled-hold-invoice` and Mostro asks the buyer for a new invoice via the `add-invoice` action. If the buyer runs **restore session** before responding, the client rebuilds the order state **only from the snapshot `status`** and shows the *"released / paying sats"* screen. The buyer is stuck with no indication that a new invoice is needed.

  Before the restore the state was correct (`payment-failed`, which offers the manual invoice input). **Restore regresses a correct state into a wrong one.**

  ### Steps to reproduce

  Requires running mostrod with the restore-with-AddInvoice support from [MostroP2P/mostro#754](https://github.com/MostroP2P/mostro/pull/754).

  1. Complete a buy until the seller releases (`active` → `fiat-sent` → release).
  2. Make Mostro's LN payout to the buyer **fail**. The order goes to `settled-hold-invoice`; after retries Mostro sends `add-invoice` asking for a new invoice.
  3. As the buyer, **don't respond** and run **Restore session**.
  4. **Actual:** the "released / paying sats" screen is shown. **Expected:** the new-invoice prompt (manual invoice input), as shown before the restore.
  
  ### Root cause

  `settled-hold-invoice` is **overloaded**: for the buyer it covers both "hold settled, sats in flight" and "payout failed, awaiting a new invoice". The protocol distinguishes them **only via the action** (`payment-failed` /  `add-invoice`) — `payment-failed` is not even a protocol `Status`, it's a client-synthesized status. So the daemon  **always** reports `settled-hold-invoice` for these orders, and the daemon re-sends `AddInvoice` to the buyer's trade key on restore ([mostro#754]https://github.com/MostroP2P/mostro/pull/754)) precisely so the client can act on it.

  The restore rebuild derives state purely from the snapshot status:

  `lib/features/restore/restore_manager.dart:519-524`

  ```dart
  case Status.settledHoldInvoice:
    return userRole == Role.buyer
        ? Action.released                    // always "paying sats"
        : Action.holdInvoicePaymentSettled;
  ```

  Applied in `restore()` (`restore_manager.dart:663-826`) via `notifier.updateStateFromMessage(...)`. Because the daemon always reports `settled-hold-invoice`, a status-based rebuild can never recover the failed-payout substate.
  The re-sent `add-invoice` messages do arrive and get saved, but they're dropped by the restore flow and ignored by the rebuild.

  Contributing factors that stop the real `add-invoice` from fixing it during restore:

  - `abstract_mostro_notifier.dart:83-87` — the `isRestoring` gate skips all message processing.
  - `order_state.dart:289-295` — `add-invoice` only keeps `paymentFailed` if the state was *already* `paymentFailed`.

  ### Proposed fix

  Have the restore rebuild use the snapshot only to seed session/role/peer, and derive the final state by replaying the stored messages (like `OrderNotifier.sync()`), so the re-sent `payment-failed`/`add-invoice` win and the order lands on `payment-failed` → manual invoice input.

  Minimal alternative: in `_getActionFromStatus`, for `settledHoldInvoice` + buyer, check stored history; if there's a `payment-failed`/`add-invoice` after the `released`, map to `Action.paymentFailed` instead of `Action.released`.

  ### Acceptance criteria

  - [ ] After restoring an order in `settled-hold-invoice` with a failed payout, the buyer sees the new-invoice
  prompt, not "paying sats".
  - [ ] The happy path (`settled-hold-invoice` with no failed payout) still shows "released / paying sats".
  - [ ] Post-restore state matches pre-restore state for this order.
  - [ ] Regression test: restore with snapshot `settled-hold-invoice` + `payment-failed`/`add-invoice` in storage →  final state `payment-failed`.

AndreaDiazCorreia:
Closes #615

When Mostro's Lightning payout to the buyer fails, the order stays at `settled-hold-invoice` and the daemon asks for a new invoice. On restore the state was rebuilt purely from the snapshot status, so this substate regressed into the "released / paying sats" screen instead of the new-invoice prompt.

`settled-hold-invoice` is overloaded for the buyer: it covers both "hold settled, sats in flight" and "payout failed, awaiting a new invoice". The snapshot status alone cannot tell them apart, so restore has to resolve the substate from the stored message history.

## Changes

- Add `restoreHasFailedPayoutSignal()`, which reads the same discriminator `OrderState.updateWith` already uses since `ca094cab`: `payment-failed`, or an `add-invoice` whose payload carries `settled-hold-invoice`. The `add-invoice` that opens the normal buyer flow carries `waiting-buyer-invoice`, so neither a partial nor an out-of-order relay-replayed history can be misread as a failed payout.
- Add `restoreReplayPlan()`, which replays `add-invoice` in place of the snapshot action for a buyer's failed payout. `updateWith` maps that to `payment-failed` on its own, so no synthetic `payment-failed` message is fabricated or persisted.
- Stamp the replayed message after every message already in storage, so a later `OrderNotifier.sync()` cannot replay an older `released` last and drag the buyer back to "paying sats".
- Hoist the session lookup that was duplicated in both branches of the restore loop.
- Add regression tests.

## Scope

`ca094cab` on `main` already covers the persistence half: `sync()` converges to `payment-failed` from the real stored `add-invoice`, whose arrival timestamp outranks the synthetic `released` stamped at `createdAt`. What remained broken, and what this PR fixes, is the **initial post-restore state**: the first render after restore, before any further message arrives.

## Acceptance criteria

- [x] Restore with a failed payout shows the new-invoice prompt, not "paying sats".
- [x] Happy path (`settled-hold-invoice` with no failed payout) still shows "released / paying sats".
- [x] A seller in `settled-hold-invoice` is unaffected.
- [x] Post-restore state matches pre-restore state and survives a later `sync()`.
- [x] Regression tests for signal detection, the replay plan (role gate, status gate, timestamp seeding) and the rebuilt state.

## Verification

- `flutter analyze`: no issues
- `flutter test`: 1116 passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order restoration for “failed payout, awaiting new invoice” cases, reliably replaying the correct recovery sequence while preserving message timing and ordering.
  * Restored order state accurately across successful and failed payout histories, including cases with missing timestamps.

* **Tests**
  * Added regression coverage for failed-payout detection, state rebuilding, replay selection, timing, ordering, and negative scenarios.

### DIFF
--- a/lib/features/restore/restore_manager.dart
+++ b/lib/features/restore/restore_manager.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:math' as math;
 import 'package:flutter/foundation.dart';
 import 'package:dart_nostr/nostr/core/key_pairs.dart';
 import 'package:dart_nostr/nostr/model/event/event.dart';
@@ -732,15 +733,15 @@ class RestoreService {
               .where((d) => d.orderId == orderDetail.id)
               .firstOrNull;
 
+          final session = ref
+              .read(sessionNotifierProvider.notifier)
+              .getSessionByOrderId(orderDetail.id);
+
           // Determine action and create dispute if needed
           Action action;
           Dispute? dispute;
 
           if (restoredDispute != null && order.status == Status.dispute) {
-            final session = ref
-                .read(sessionNotifierProvider.notifier)
-                .getSessionByOrderId(orderDetail.id);
-
             final initiator = restoredDispute.initiator;
             final bool userInitiated =
                 (initiator == 'buyer' && session?.role == Role.buyer) ||
@@ -766,9 +767,6 @@ class RestoreService {
             logger.i('Restore: dispute found for order ${orderDetail.id}');
           } else {
             // Regular order without dispute
-            final session = ref
-                .read(sessionNotifierProvider.notifier)
-                .getSessionByOrderId(orderDetail.id);
             action = _getActionFromStatus(order.status, session?.role);
           }
 
@@ -838,19 +836,38 @@ class RestoreService {
               }
             }
 
+            // `settled-hold-invoice` is overloaded for the buyer (see issue
+            // #615): it covers both "hold settled, sats in flight" and "payout
+            // failed, awaiting a new invoice", so rebuilding from the snapshot
+            // status alone always lands on the "paying sats" screen. Resolve
+            // the substate from the stored history instead.
+            final replay = await restoreReplayPlan(
+              snapshotAction: action,
+              snapshotTimestamp: orderDetail.createdAt ??
+                  DateTime.now().millisecondsSinceEpoch,
+              status: order.status,
+              role: session?.role,
+              loadStoredMessages: () =>
+                  storage.getAllMessagesForOrderId(orderDetail.id),
+            );
+            if (replay.action != action) {
+              logger.i(
+                'Restore: detected failed payout for order ${orderDetail.id}, '
+                'replaying ${replay.action.value} instead of ${action.value}',
+              );
+            }
+
             // Create regular order message with Order payload
             final mostroMessage = MostroMessage<Order>(
               id: orderDetail.id,
-              action: action,
+              action: replay.action,
               payload: order,
-              timestamp:
-                  orderDetail.createdAt ??
-                  DateTime.now().millisecondsSinceEpoch,
+              timestamp: replay.timestamp,
             );
 
             // Save order message to storage
             final key =
-                '${orderDetail.id}_restore_${action.value}_${DateTime.now().millisecondsSinceEpoch}';
+                '${orderDetail.id}_restore_${replay.action.value}_${DateTime.now().millisecondsSinceEpoch}';
             await storage.addMessage(key, mostroMessage);
 
             // Update state with order message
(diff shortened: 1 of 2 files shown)


grunch (changes requested):
## Review: fix(restore): show new-invoice prompt for failed payout in settled-hold-invoice

Verified locally on a `main` + this branch merge (clean merge): `flutter analyze` → no issues, `flutter test test/features/restore test/features/order` → 119 passed. The problem is not the code quality of the diff, it is the premise the heuristic is built on and how it interacts with what has landed on `main` since this branch was cut.

### 🔴 HIGH — The `mostro#754` invariant this PR relies on is not on the daemon's `main`

`restoreHasFailedPayoutSignal()`'s doc comment, the inline comment in `restore()` and the `lone re-sent add-invoice is a failed-payout signal (mostro#754)` test all state as an invariant that "on `restore-session` the daemon re-enqueues ONLY `add-invoice` … for orders flagged `failed_payment = true`".

- MostroP2P/mostro#754 was **closed unmerged** on 2026-08-20 (`merged_at: null`).
- `src/app/restore_session.rs` on mostro `main` today only calls `send_restore_session_response` / `send_restore_session_timeout`; there is no `AddInvoice` re-send at all.

So against a production daemon, the only way a signal reaches storage is relay-replayed gift-wrap history during the bounded 10 s window (`restore_manager.dart:701`). That is exactly the partial-history scenario Codex and ermeme's first review flagged: a happy-path buyer whose relay delivered the early `add-invoice` (waiting-buyer-invoice) but not yet `released` is rebuilt as `payment-failed`. ermeme's later approval explicitly withdrew the concern *because of* the #754 daemon behaviour, which is not in effect. Please don't encode an unmerged daemon PR as a client invariant.

**Suggested fix** — drop the ordering heuristic entirely and use the discriminator `main` already relies on (`ca094cab`, "detect the payout add-invoice from the message payload"): the payout `add-invoice` carries an `Order` payload with `status == settled-hold-invoice`, the early one carries `waiting-buyer-invoice`. That is unambiguous regardless of relay completeness or ordering, and needs no `sort`:

```dart
bool restoreHasFailedPayoutSignal(List<MostroMessage> messages) =>
    messages.any(
      (m) =>
          m.action == Action.paymentFailed ||
          (m.action == Action.addInvoice &&
              m.getPayload<Order>()?.status == Status.settledHoldInvoice),
    );
```

Then rewrite the doc comment/tests around the payload, not around #754.

### 🔴 HIGH — Branch is 56 commits behind `main`; `main` already changed this exact flow

Since this branch was cut, `main` merged `ca094cab`, `8dc4004c` and `2813313a`, which all touch `Action.addInvoice` / `Status.settledHoldInvoice` / `Status.paymentFailed` in `OrderState`. Consequences worth re-checking after a rebase:

- With `ca094cab`, `OrderNotifier.sync()` already converges to `paymentFailed` for the stored real `add-invoice` (its timestamp is the arrival time, later than the synthetic `released` stamped at `createdAt`). The PR is still needed for the *initial* post-restore state, but the description/comments should reflect that the persistence half is now covered by `main` and this change is about the first render.
- Replaying only `[Action.addInvoice]` (with the `settled-hold-invoice` payload) now yields `Status.paymentFailed` + `Action.addInvoice` on `main` on its own — see next point on why replaying a synthetic `payment-failed` is undesirable.

### 🟡 MEDIUM — Synthetic `payment-failed` message is persisted with a bogus payload (`restore_manager.dart:864-903`)

The replayed `Action.paymentFailed` carries the `Order` payload, so `OrderState.paymentFailed` stays `null`, and the message is written to storage as if Mostro sent it. `mostro_message_detail_widget.dart:284-290` renders that action from the `PaymentFailed` payload → the user sees "payment failed, 0 attempts, retry every 0 min". Prefer replaying only `add-invoice` (see above), or at minimum don't persist a fabricated `payment-failed` event.

### 🟡 MEDIUM — Tests don't cover the `restore()` branch that changed

`restore_failed_payout_test.dart` only exercises the pure helper and `OrderState.updateWith`. Nothing covers:
- the role gate (a **seller** in `settled-hold-invoice` must stay on `released`/`hold-invoice-payment-settled`),
- the timestamp seeding above `max(stored timestamps)` — i.e. the fix in the second commit, which Codex flagged and which is the reason the fix survives `sync()` — is untested,
- the pair actually being persisted and staggered (`baseTimestamp + i`).

Also `returns false for an early add-invoice that precedes the release` builds the "early" `add-invoice` with a `settled-hold-invoice` payload, which the daemon never does, so it validates ordering rather than the real signal. A `restore_manager_dispute_test.dart`-style test with a fake storage would cover the branch.

### 🔵 LOW

- `restore_manager.dart:1249-1255`: `Action.release` is buyer→never (seller sends it) and `Action.holdInvoicePaymentSettled` is delivered to the seller; this branch is gated on `Role.buyer`, so only `Action.released` can ever match. Dead conditions that mislead readers — they disappear with the payload-based check.
- `restore_manager.dart:872-876`: `messages.fold(... ? m.timestamp! : max)` — simpler and null-safe without `!`: `messages.map((m) => m.timestamp ?? 0).fold(baseTimestamp, math.max) + 1`.
- `restore_manager.dart:849-851`: third `getSessionByOrderId(orderDetail.id)` lookup in the same loop iteration (`session` is already read at 736-738 / 771-773); hoist one lookup to the top of the loop body.
- The two consecutive comment blocks (`Seed replay timestamps from the order creation time…` vs `Stamp the replayed pair after every message already in storage…`) contradict each other; keep one that describes the actual rule.
- Doc comment says "`payment-failed` is a client-synthesized status that never travels on the wire" — the *status* is client-side, but `Action.paymentFailed` is sent by the daemon on the first failure (`ca094cab` commit message). Worth rewording so nobody removes the `paymentFailed` check believing it is unreachable.

### Verdict
Requesting changes for the two HIGH items: rebase on `main` and replace the #754-based ordering heuristic with the payload-status discriminator. The acceptance criteria in the description are otherwise sound, and the money-at-risk bug (#615) is worth fixing.

Catrya:
It was decided to always display the add-invoice button, so it is no longer necessary to receive add-invoices from the restore. This PR is closed in favor of #667

PR #624, opened 2026-06-19
feat: add dual send for transport v2 (Phase B)

AndreaDiazCorreia:
Implements **Phase B — Dual send** of the transport v1 → v2 migration (`docs/architecture/TRANSPORT_V2_MIGRATION.md`). Builds on Phase A (#621, merged), routing outbound Mostro messages through the per-node resolved transport.

## Changes
- **v2 wrap** (`MostroMessage.wrapNip44`): builds the `[message, tradeSig, identityProof]` 3-tuple with `version: 2`, NIP-44 encrypts it toward the node, and emits a kind-14 event **signed by the trade key** with a `p` tag. PoW is preserved (mined on the kind-14 id).
- **Identity proof** mirrors `mostro-core`'s `transport.rs`: trade-key signature = existing `MostroMessage.sign` (SHA-256 hex digest, Schnorr); identity proof = master key signing `mostro-transport-v2-identity:<tradePubkey>:<messageJSON>`. Both are `null` in full-privacy mode.
- **`version` threading**: derived from the resolved transport instead of the global constant (1 for gift wrap, 2 for NIP-44).
- **Routing** (`MostroService.publishOrder`): selects `wrapNip44` (v2) or `wrap` (v1) from the node's `protocol_version`.

## Scope
- The v1 gift-wrap send path is **byte-for-byte unchanged**; v2 send only engages for nodes advertising `protocol_version=2` (none in production yet).
- The NIP-40 `expiration` tag is omitted (optional/caller-supplied per the daemon), avoiding any risk of a message expiring before processing.
- Out of scope: send-side wiring polish (Phase C), additional tests (Phase D).

## Verification
- `flutter test`: all pass, including a new v2 wrap → unwrap round-trip (reputation + full-privacy) that cryptographically verifies the trade signature and identity proof, plus v1 regression.

## Summary by CodeRabbit

* **New Features**
  * Added NIP-44 encrypted direct messaging for protocol v2 with protocol-versioned signing/serialization.
  * Outbound message wrapping now routes per the connected Mostro node’s protocol version, covering restore and dispute flows.
  * Restore receiving now supports both v1 gift-wrap and v2 direct responses.
* **Documentation**
  * Updated transport migration and session recovery docs to reflect dual receive/send behavior.
* **Tests**
  * Added regression tests for NIP-44 wrapping and restore decoding across both transports, including signature validation cases.

### DIFF
--- a/docs/architecture/SESSION_RECOVERY_ARCHITECTURE.md
+++ b/docs/architecture/SESSION_RECOVERY_ARCHITECTURE.md
@@ -151,7 +151,13 @@ Future<void> importMnemonicAndRestore(String mnemonic) async {
 
 **File**: `lib/features/restore/restore_manager.dart:116-141`
 
-Creates a temporary subscription using trade key index 1 to receive Mostro responses:
+Creates a temporary subscription using trade key index 1 to receive Mostro
+responses. To interoperate across the transport v2 migration it listens on
+**both** wire transports at once — v1 gift wrap (kind 1059) and v2 NIP-44 direct
+(kind 14) — because the node info (kind 38385) that advertises `protocol_version`
+may not have loaded yet when restore starts. The node answers on whichever
+transport it speaks; the v2 filter pins `authors = [mostroPubkey]` to disambiguate
+from NIP-17 chat (also kind 14). See `TRANSPORT_V2_MIGRATION.md`.
 
 ```dart
 Future<StreamSubscription<NostrEvent>> _createTempSubscription() async {
@@ -159,18 +165,30 @@ Future<StreamSubscription<NostrEvent>> _createTempSubscription() async {
     throw Exception('Temp trade key not initialized');
   }
 
-  final filter = NostrFilter(
+  final mostroPubkey = ref.read(settingsProvider).mostroPublicKey;
+  final v1Filter = NostrFilter(
     kinds: [1059],
     p: [_tempTradeKey!.public],
     limit: 0, // No historical events, only new ones
   );
+  final v2Filter = NostrFilter(
+    kinds: [14],
+    authors: [mostroPubkey],
+    p: [_tempTradeKey!.public],
+    limit: 0,
+  );
 
-  final request = NostrRequest(filters: [filter]);
+  final request = NostrRequest(filters: [v1Filter, v2Filter]);
   final stream = ref.read(nostrServiceProvider).subscribeToEvents(request);
-  
+
   return stream.listen(_handleTempSubscriptionsResponse);
 }
 ```
+
+The response is decoded by the top-level `decodeRestoreMessage(event, tempTradeKey,
+mostroPubkey)`, which branches on `event.kind`: kind 14 is NIP-44 decrypted (and
+the node signature verified) straight to the tuple, while kind 1059 is gift-wrap
+unwrapped to a rumor whose content is the tuple. Both converge on `tuple[0]`.
 
 ### Stage 3: Data Request Sequence
 
--- a/docs/architecture/TRANSPORT_V2_MIGRATION.md
+++ b/docs/architecture/TRANSPORT_V2_MIGRATION.md
@@ -304,11 +304,15 @@ unchanged.
   `version: 2`; computes the trade signature; computes the identity proof
   (domain-tagged string signed with the master key, `null` in full-privacy);
   NIP-44 encrypts the 3-tuple toward the node; emits a kind-`14` event **signed
-  by the trade key** with `p` and NIP-40 `expiration` tags.
-- Route `MostroService.publishOrder` through the resolved transport (reuses the
-  Phase A resolver / `protocol_version`). **Preserve PoW**
-  (`NostrUtils.mineProofOfWork`) for the first-contact lane — the daemon may
-  still require PoW on the kind-14 event id.
+  by the trade key** with a `p` tag (the NIP-40 `expiration` tag is omitted; see
+  the note below).
+- Route **every** outbound Mostro send through the resolved transport via a
+  single `MostroMessage.wrapForTransport(protocolVersion: …)` entry point — not
+  just `MostroService.publishOrder`, but also the `RestoreManager` requests
+  (restore, order-details, last-trade-index) and
+  `DisputeRepository.createDispute`, so a v2 node never receives a stray v1 gift
+  wrap. **Preserve PoW** (`NostrUtils.mineProofOfWork`) for the first-contact
+  lane — the daemon may still require PoW on the kind-14 event id.
 - **Identity proof signature** mirrors `mostro-core`'s `transport.rs`: the
   trade-key signature (tuple element 1) is the existing `MostroMessage.sign`
   (SHA-256 hex digest then Schnorr), and the identity proof (element 2) is the
--- a/docs/architecture/TRANSPORT_V2_MIGRATION.md
+++ b/docs/architecture/TRANSPORT_V2_MIGRATION.md
@@ -18,10 +18,10 @@ during the migration window.
 > - **Reference client (CLI)**: `MostroP2P/mostro-cli` PRs #176, #177, #178 and
 >   its `docs/TRANSPORT_V2_SPEC.md`.
 >
-> **Status.** Living design specification. **Phase A (dual receive) is
-> implemented** in this branch, including `protocol_version` auto-detection and
-> per-node transport resolution on the receive path. The remaining phases (§5)
-> are pending.
+> **Status.** Living design specification. **Phase A (dual receive)** —
+> including `protocol_version` auto-detection and per-node transport resolution
+> on the receive path — is merged to `main`. **Phase B (dual send)** is
+> implemented in this branch. Phases C–D (§5) are pending.
 
 ---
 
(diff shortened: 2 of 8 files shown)


grunch (commented):
## Strict Review — Phase B (Transport v2, Dual send)

**Verdict: Approvable with minor changes.** The architecture is solid, well-documented, and well-tested (the tests cryptographically verify the trade signature and identity proof, not just the tuple shape). I confirmed locally that `flutter analyze` reports 0 issues and all 6 new tests pass. ermeme's blocking concern (restore only received kind 1059) is **resolved** by the dual-receive commits, and the doc contradiction at line 307 (NIP-40 `expiration` tag) is **already fixed**.

### Status of prior feedback
- ✅ **Codex P2** (unconverted direct sends): resolved — there are **no remaining direct `.wrap()` sends** in `lib/`; everything routes through `wrapForTransport`.
- ✅ **ermeme** (restore receive path): resolved — `_createTempSubscription` now subscribes to both kind 1059 and kind 14 (the latter scoped with `authors:[mostroPubkey]`, good).
- ⚠️ **CodeRabbit Major** (dispute missing `masterKey`/`keyIndex`): **still present and valid** — see below.

### Findings

**1. (Medium) `DisputeRepository.createDispute` omits `masterKey`/`keyIndex` → in v2 reputation mode the dispute is sent without identity proof**
`dispute_repository.dart:55-60` calls `wrapForTransport` without `masterKey`/`keyIndex`. This is the **actively used** path (`trade_detail_screen.dart:774`). Consequence: in `wrapNip44`, when `masterKey == null`, **both `tradeSig` and `identityProof` are `null`** → a v2 node receives a dispute with no identity binding in reputation mode. This is inconsistent with `publishOrder` (`mostro_service.dart:372-373`), which does pass them. In v1 this is pre-existing behavior (the seal is signed by the trade key), but the PR's own docs state this path is a first-class v2 citizen, so it's worth closing. `session` already exposes `masterKey`/`keyIndex`/`fullPrivacy`, so the fix is trivial:

```dart
masterKey: session.fullPrivacy ? null : session.masterKey,
keyIndex: session.fullPrivacy ? null : session.keyIndex,
```

**2. (Low) `decodeRestoreMessage` — missing bounds check before `contentList[0]`** (`restore_manager.dart:~1098`)
It validates `content.isEmpty` but not `contentList.isEmpty`. An empty JSON array `[]` would throw `RangeError` instead of a controlled exception. Defensive fix:

```dart
if (contentList.isEmpty) throw Exception('Restore message tuple is empty');
```

**3. (Nit) Stale comments** in `dispute_repository.dart:43,46` still say "Gift Wrap protocol (NIP-59)" even though the path can now emit kind 14.

**4. (Verify before enabling v2 — non-blocking) Interop with `mostro-core`**
`tradeSig` is computed over `jsonEncode(messageMap)` and the identity proof over `mostro-transport-v2-identity:<tradePubkey>:<messageJSON>`. This relies on the JSON key ordering matching `transport.rs`'s re-serialization on the daemon byte-for-byte. Since it reuses the same `toJson()` as the v1 `sign()` (which already works in production), it should match, but **there is no v2 node in production yet**, so this is not validated against a real daemon. Recommend an end-to-end interop test before activating v2.

**5. (Minor note) `wrapNip44` uses `createdAt: DateTime.now()`** with no timestamp randomization. Acceptable since the kind-14 event is signed by the trade key (already linkable), but worth being aware of.

### Conclusion
None of these are blocking for the phase (no v2 nodes in production). I recommend **addressing #1 before merge** (identity consistency + it's the active path), with #2/#3 as quick cleanup. #4 is a pre-launch to-do for v2.

grunch (commented):
## Follow-up: wire-format verification against `mostro-core/src/transport.rs`

Re-framing my earlier review per a fair point: **there are no v2 nodes in production *because* there is no v2 client yet — this PR is the prerequisite.** So the wire format isn't a "verify later" item; it's the contract that future v2 daemons will be built against. I verified `wrapNip44` byte-for-byte against the reference impl (`MostroP2P/mostro-core:src/transport.rs::wrap_message_nip44` + `Message::sign`). **It matches — this implementation is spec-compliant.** Details, so reviewers don't have to re-derive it:

| Aspect | mostro-core reference | This PR (`wrapNip44`) | Match |
|---|---|---|---|
| Event kind | `Kind::PrivateDirectMessage` (14) | `kind: 14` | ✅ |
| Tuple | `(Message, Option<trade_sig>, Option<(id_pub, id_sig)>)` → JSON array | `[messageMap, tradeSig, identityProof]` | ✅ |
| Tuple element 0 | `message.as_json()` = `{"order"/"restore": {...}}` | `{wrapperKey: toJson(version:2)}` | ✅ |
| Sign scheme | `schnorr(sha256_digest_raw_32B)` | `sign(hex(sha256(msg)))` → dart_nostr hex-decodes to the same 32B | ✅ |
| Identity domain | `mostro-transport-v2-identity:<trade_pubkey_hex>:<message_json>` | `mostro-transport-v2-identity:${tradeKey.public}:$messageJson` | ✅ |
| identityProof shape | `[identity_pubkey_hex, sig]` or `null` | `[masterKey.public, sig]` or `null` | ✅ |
| NIP-44 keys | `encrypt(trade_secret, receiver)` | `encryptNIP44(tuple, tradeKey.private, recipient)` | ✅ |
| Author / event sig | signed by trade key | `keyPairs: tradeKey` | ✅ |
| `p` tag | `Tag::public_key(receiver)` | `['p', recipientPubKey]` | ✅ |
| PoW | `EventBuilder.pow(opts.pow)`, re-signed | `mineProofOfWork(event, difficulty, tradeKey)` | ✅ |
| `created_at` | builder default (real time, no randomization) | `DateTime.now()` | ✅ (retracting my earlier note — the reference doesn't randomize either) |
| NIP-40 expiration | optional (`opts.expiration`) | omitted | ✅ (valid: daemon makes it optional) |

The one cross-impl risk — that Dart's `jsonEncode` key ordering must equal the daemon's serde `message.as_json()` for `tradeSig`/identity-proof to verify — is **already proven by the v1 path in production**, which signs the exact same `{wrapperKey: toJson()}` and is verified the same way (`Message::verify_signature` over `message.as_json()`). v2 reuses both sides verbatim. So no new canonicalization risk.

### One real consequence to fix before merge (sharpening finding #1)

Given the daemon's `unwrap_message_nip44`, a `null` `identity_proof` is **accepted** and sets `identity = event.pubkey` (the trade key) — i.e. it is treated as **full-privacy mode**. Because `DisputeRepository.createDispute` omits `masterKey`/`keyIndex`, a dispute raised by a **reputation-mode** user on a v2 node would be silently received as full-privacy: the daemon binds it to the trade key, not the user's master identity. Same omission exists in the v1 path here (pre-existing), but since we're establishing the v2 contract strictly, this is the moment to fix it so reputation disputes carry the identity proof. Fix is the 2 lines I noted (mirror `publishOrder`: `masterKey`/`keyIndex` gated on `session.fullPrivacy`).

### Net
Crypto/serialization interop: **verified correct**. Remaining actionable items unchanged: dispute identity proof (#1, now with a concrete consequence), `contentList` empty-array guard (#2), stale NIP-59 comments (#3). Recommend a real end-to-end roundtrip test against a `mostro-core` v2 node (or a vector generated by `wrap_message_nip44`) added in Phase D before any v2 node ships.

Reference: [`mostro-core/src/transport.rs`](https://github.com/MostroP2P/mostro-core/blob/main/src/transport.rs), [`TRANSPORT_V2_SPEC.md`](https://github.com/MostroP2P/mostro/blob/main/docs/TRANSPORT_V2_SPEC.md)

grunch (changes requested):
## Request Changes

The wire format is verified spec-compliant against `mostro-core/src/transport.rs` (see my detailed comparison comment), and the dual-receive regression is resolved. Approving is blocked only by the following, which should be addressed before merge:

**Must fix**
1. `DisputeRepository.createDispute` omits `masterKey`/`keyIndex`, so a reputation-mode dispute on a v2 node is silently downgraded to full-privacy identity (daemon accepts `null` identity_proof and binds the dispute to the trade key, not the master identity). Mirror `publishOrder`:
   ```dart
   masterKey: session.fullPrivacy ? null : session.masterKey,
   keyIndex: session.fullPrivacy ? null : session.keyIndex,
   ```

**Should fix**
2. `decodeRestoreMessage`: add an empty-array guard before `contentList[0]` to avoid `RangeError` on a `[]` payload.
3. Stale comments in `dispute_repository.dart:43,46` still say "Gift Wrap protocol (NIP-59)" though the path can now emit kind 14.

**Recommended (Phase D)**
- Add an end-to-end roundtrip test against a `mostro-core` v2 wrap vector before any v2 node ships.

Once #1–#3 land I'm happy to approve.

grunch (approved):
## Approve

Must/should-fix items from the previous review are addressed in `e7bbcfdf`:

1. ✅ `DisputeRepository.createDispute` now passes `masterKey`/`keyIndex` (gated on `session.fullPrivacy`), so reputation-mode v2 disputes carry the identity proof instead of being downgraded to full-privacy.
2. ✅ `decodeRestoreMessage` guards against an empty tuple before indexing `[0]`.
3. ✅ Stale NIP-59-only comments in the dispute send path refreshed.

Verified locally: `flutter analyze` clean, nip44 + restore-decode tests pass. Wire format remains spec-compliant with `mostro-core/src/transport.rs`.

Non-blocking follow-up for Phase D: add an end-to-end roundtrip test against a `mostro-core` v2 wrap vector before any v2 node ships.

PR #628, opened 2026-06-23
revert: roll back Riverpod 3.x migration (#613) and dependent PRs (#624, #625)

grunch:
## Summary
Reverts the Riverpod 3.x migration (#613), which broke the app, and the two PRs merged on top of it (#624, #625). This restores `main` to the exact state of #619.

## Reverted commits (LIFO order)
- Revert #625 — `feat: add MWK (Malawi Kwacha) payment methods`
- Revert #624 — `feat: add dual send for transport v2 (Phase B)`
- Revert #613 — `feat: migrate to Riverpod 3.x`

## Verification
- Reverts applied cleanly, no conflicts.
- Resulting tree is **identical** to #619 (`git diff fe3a897f HEAD` is empty).
- Returns the project to Flutter 3.35.7 + Riverpod 2.6.1 + analyzer 7.7.1.

## Follow-up
The content of #624 and #625 is re-applied — **without any Riverpod 3.x changes** — in two separate PRs that stack on top of this one:
- transport v2 Phase B → `feat/transport-v2-phase-b-reapply`
- MWK payment methods → `feat/mwk-payment-methods-reapply`

**Merge this PR first.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified message transport protocol for improved reliability.
  * Updated session recovery mechanism to use optimized encryption layer.

* **Chores**
  * Updated Flutter, Gradle, and dependency versions for compatibility.
  * Migrated to modern icon library package.
  * Refactored state management framework usage.
  * Removed MWK from supported payment methods.

PR #631, opened 2026-06-23
fix: resolve provider init and node-switch restore connectivity races

AndreaDiazCorreia:
## Summary
Fixes two pre-existing race conditions that surface after the Riverpod 3.x rollback (#628), both caused by code that the revert does not touch:

1. `Exception: Nostr is not initialized. Call init() first.` during app start, logged as "Failed to init mostro instance listener".
2. `TimeoutException: Stage RestoreStage.gettingRestoreData timed out after 10s` when switching Mostro nodes.

Neither is introduced by the revert; the Riverpod 2.x timing change merely exposes them.

### 1. Order repository init race
`OpenOrdersRepository`'s constructor eagerly calls `subscribeToEvents()`, which throws when `NostrService` is not yet initialized. The transport Phase A listener (`SubscriptionManager._initMostroInstanceListener`) forces the first build of `orderRepositoryProvider` during `SubscriptionManager` construction, which can run before `init()` completes. The thrown error is cached on the provider, poisoning every later read of the order repository.

### 2. Node-switch restore timeout
Since the move to bootstrap relay discovery (#610), no relays are seeded and the selected node's relays are rediscovered asynchronously via kind 10002. On node switch, `initRestoreProcess()` sends the restore request immediately, while the
new node's relays take up to ~5s (periodic resync) plus discovery to connect, so the request races connectivity and times out before any relay can route it.

## Changes
- `OpenOrdersRepository._subscribeToOrders()`: defer the subscription with a bounded retry when `NostrService` is not initialized, instead of throwing and poisoning the provider.
- `OpenOrdersRepository.updateSettings()`: reset `_mostroInstance` on instance change so stale node info is not reported for the new instance.
- `RestoreService`: add `_waitForNodeConnectivity()`, called before sending the restore request. It engages the bootstrap relays immediately (instead of waiting for the periodic watchdog) and waits for the node's info event (kind 38385) within the discovery window, falling back to a best-effort attempt with a live-relay floor.

## Verification
- `flutter test`: all tests pass (473).
- Runtime: confirm node switching no longer times out for default and custom
  nodes, and the order book loads cleanly on cold start.

## Notes
- Stacks on top of #628 (Riverpod rollback). Base this PR on
  `revert/riverpod-3x-rollback` (or rebase onto `main` once #628 is merged).

## Release Notes

* **New Features**
  * Improved restore mode state management and progress flow.
  * Enhanced navigation notification indicators for chats and order book.

* **Bug Fixes**
  * Improved session recovery by waiting for node connectivity before starting restores.
  * Improved startup reliability by delaying order subscriptions until initialization is ready (with bounded retries).
  * Simplified restore message decoding to a single gift-wrap path to reduce transport-related edge cases.

* **Refactor**
  * Streamlined message signing/wrapping and unified wrapping behavior across ordering and disputes.

* **Chores**
  * Updated Flutter/Gradle/Kotlin toolchain versions; upgraded dependencies and icon packages.
  * Removed obsolete NIP-44/NIP-59 v2 transport test coverage; cleaned up unused/legacy Riverpod imports.
  * Updated payment methods by removing the MWK currency entry.

### DIFF
--- a/lib/data/repositories/open_orders_repository.dart
+++ b/lib/data/repositories/open_orders_repository.dart
@@ -29,6 +29,13 @@ class OpenOrdersRepository implements OrderRepository<NostrEvent> {
   final Map<String, NostrEvent> _events = {};
   StreamSubscription<NostrEvent>? _subscription;
 
+  /// Polls for NostrService readiness when the repository is built before
+  /// init() completes, so the order subscription can be opened once Nostr is up.
+  Timer? _initRetryTimer;
+
+  static const _initRetryInterval = Duration(milliseconds: 200);
+  static const _maxInitRetries = 150; // ~30s before giving up
+
   NostrEvent? get mostroInstance => _mostroInstance;
 
   Stream<NostrEvent> get mostroInstanceStream =>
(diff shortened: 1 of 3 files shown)


Catrya (dismissed):
tACK

Issue #635, opened 2026-06-28
Submit button doesn't work if the keyboard is on the screen.


wchancao:
**Describe the bug**
Submit button doesn't work if the keyboard is on the screen.

**To Reproduce**
1. Take a buy order
2. Input the invoice
3. Clicking the submit button does nothing if the keyboard is on the screen

**Expected behavior**
Clicking the submit button should send the invoice

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/923c0c7f-57d1-482a-8f64-ca22ded042bb" />

**Smartphone (please complete the following information):**
 - Device: Google Pixel 8a
 - OS: Android 16
 - Mostro v1.2.9

codaMW:
Traced this to the invoice form overflowing when the keyboard shrinks the viewport, which clips the Submit button's tap area. Opened #636 with a fix (SingleChildScrollView wrap). It didn't reproduce on my own device (viewport-specific), so @wchancao could you confirm it resolves on your Pixel 8a?

wchancao:
Same behavior in v1.3.0, but not sure if the fix was included in that version

PR #636, opened 2026-07-01
fix(invoice): keep Submit reachable on Add Lightning Invoice when keyboard is open (#635)

codaMW:
> ## What
> Wraps the invoice form in `add_lightning_invoice_widget.dart` in a `SingleChildScrollView` so it scrolls instead of overflowing when the soft keyboard is open.
>
> ## Root cause
> `build` returned a bare `Column` containing a 6-line invoice `TextFormField`. When the keyboard opens, the `Scaffold` shrinks the available height; on some screen geometries the `Column` overflows and the Cancel/Submit `Row` is pushed past the hittable area the button renders but its tap target is clipped, so Submit does nothing. Dismissing the keyboard restores the viewport and it works again.
>
> ## Fix
> Wrapping the `Column` in a `SingleChildScrollView` keeps Submit reachable on any screen size. The `-w` diff shows the real change is just the wrapper; the rest is `dart format` reflow.
>
> ## Testing
> - `flutter analyze`: no issues.
> - Reached the Add Lightning Invoice screen via a full local E2E, but **the bug did not reproduce on my device (Nokia C31)**  Submit worked with the keyboard open, and still worked when I forced a smaller viewport via split-screen. Consistent with the diagnosis: the overflow only occurs on certain screen geometries. The report is on a Pixel 8a / Android 16.
> - Verified no-regression on my C31: Submit still works and the form now scrolls with the keyboard open.
> - @wchancao, could you confirm this resolves it on your Pixel 8a? I couldn't trigger the original overflow locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * The Lightning invoice entry screen now scrolls, making it easier to use on smaller displays or when the keyboard is open.
  * Existing invoice input, cancel, and submit actions remain unchanged.

### DIFF
--- a/lib/shared/widgets/add_lightning_invoice_widget.dart
+++ b/lib/shared/widgets/add_lightning_invoice_widget.dart
@@ -30,89 +30,93 @@ class AddLightningInvoiceWidget extends StatefulWidget {
 class _AddLightningInvoiceWidgetState extends State<AddLightningInvoiceWidget> {
   @override
   Widget build(BuildContext context) {
-    return Column(
-      crossAxisAlignment: CrossAxisAlignment.start,
-      children: [
-        Text(
-          S.of(context)!.pleaseEnterLightningInvoiceFor(
-            widget.amount.toString(),
-            widget.fiatCode,
-            widget.fiatAmount,
-            widget.orderId,
-          ),
-          style: const TextStyle(
-            color: AppTheme.textPrimary,
-            fontSize: 16,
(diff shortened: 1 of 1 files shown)


Catrya (approved):
tACK good job @codaMW

BraCR10 (approved):
tACK , thanks @codaMW 

Tested on:
- Samsung Galaxy A14 
- Pixel 7 emulator

PR #655, opened 2026-08-17
fix: add value equality to Order

### LINKED ISSUE #652: Order lacks value equality, so OrderState/Dispute comparisons always report a change
## Context

Found while writing tests for `OrderState` (PR #651).

`lib/data/models/order.dart` declares no `operator ==` and no `hashCode`, so `Order` falls back to identity comparison. Every sibling payload model does define value equality — `PaymentRequest` (`payment_request.dart:93`), `Dispute` (`dispute.dart:298`), `Peer`, `CantDo`, `Amount`, `RatingUser`, `TextMessage`, `RangeAmount`, `Currency`.

## Why it matters

`OrderState.==` and `OrderState.hashCode` both include `order`:

```dart
// lib/features/order/models/order_state.dart
other is OrderState &&
    other.status == status &&
    other.action == action &&
    other.order == order &&   // <- identity comparison
    ...
```

`OrderState` is Riverpod state. Riverpod skips notifying listeners when the new state `==` the old one. Because two structurally identical `Order` instances are never equal, **any** `OrderState` carrying an order compares unequal to itself, so every rebuild of that state notifies every listener and re-renders the trade UI even when nothing changed.

`Dispute.==` and `Dispute.hashCode` have the same problem: both include `order`, so two otherwise-identical disputes never compare equal once an order is attached.

## Reproduction

```dart
Order build() => const Order(
      id: 'order-1',
      kind: OrderType.sell,
      status: Status.pending,
      amount: 50000,
      fiatCode: 'USD',
      fiatAmount: 100,
      paymentMethod: 'Wire transfer',
    );

final a = OrderState(status: Status.pending, action: Action.newOrder, order: build());
final b = OrderState(status: Status.pending, action: Action.newOrder, order: build());

print(a == b); // false — expected true
```

The current behaviour is pinned in `test/features/order/models/order_state_test.dart`:

```dart
test('states holding equal-but-distinct orders compare unequal today', () {
  expect(baseState(), isNot(baseState()));
});
```

## How to fix

Add value equality to `Order`, matching the style already used by `PaymentRequest` and `Dispute`:

```dart
@override
bool operator ==(Object other) {
  if (identical(this, other)) return true;
  return other is Order &&
      other.id == id &&
      other.kind == kind &&
      other.status == status &&
      other.amount == amount &&
      other.fiatCode == fiatCode &&
      other.minAmount == minAmount &&
      other.maxAmount == maxAmount &&
      other.fiatAmount == fiatAmount &&
      other.paymentMethod == paymentMethod &&
      other.premium == premium &&
      other.masterBuyerPubkey == masterBuyerPubkey &&
      other.masterSellerPubkey == masterSellerPubkey &&
      other.buyerTradePubkey == buyerTradePubkey &&
      other.sellerTradePubkey == sellerTradePubkey &&
      other.buyerInvoice == buyerInvoice &&
      other.createdAt == createdAt &&
      other.expiresAt == expiresAt;
}

@override
int get hashCode => Object.hashAll([
      id, kind, status, amount, fiatCode, minAmount, maxAmount,
      fiatAmount, paymentMethod, premium, masterBuyerPubkey,
      masterSellerPubkey, buyerTradePubkey, sellerTradePubkey,
      buyerInvoice, createdAt, expiresAt,
    ]);
```

`Order` is already immutable with a `const` constructor, so this is safe.

Then flip the pinned test in `order_state_test.dart` from `isNot` to an equality assertion, and re-check the `OrderState`/`Dispute` equality tests.

## Watch out for

Anything that relies on the current identity semantics to force a rebuild — grep for `orderNotifierProvider`, `ref.listen<OrderState>` and `state = ` in `lib/features/order/notifiers/`. After the fix a genuinely unchanged state stops notifying, which is the intent, but code that mutated something *outside* the compared fields and relied on the notification would need to be updated.

grunch:
Closes #652

## Problem

`lib/data/models/order.dart` declared no `operator ==` and no `hashCode`, so `Order` fell back to identity comparison — while every sibling payload model (`PaymentRequest`, `Dispute`, `Peer`, `CantDo`, `Amount`, `RatingUser`, `TextMessage`, `RangeAmount`, `Currency`) defines value equality.

`OrderState.==` and `OrderState.hashCode` both include `order`. Since two structurally identical `Order` instances were never equal, **any** `OrderState` carrying an order compared unequal to itself. `OrderState` is Riverpod state, and Riverpod skips notifying listeners only when the new state `==` the old one — so every rebuild notified every listener and re-rendered the trade UI even when nothing had changed.

`Dispute.==`/`Dispute.hashCode` also include `order`, so two otherwise-identical disputes never compared equal once an order was attached.

## Change

- `Order` gains `operator ==` and `hashCode` over all 17 fields (`id`, `kind`, `status`, `amount`, `fiatCode`, `minAmount`, `maxAmount`, `fiatAmount`, `paymentMethod`, `premium`, `masterBuyerPubkey`, `masterSellerPubkey`, `buyerTradePubkey`, `sellerTradePubkey`, `buyerInvoice`, `createdAt`, `expiresAt`), matching the style already used by `PaymentRequest` and `Dispute`. `Order` is immutable with a `const` constructor, so this is safe.

No other production code changed.

## Tests

- `test/models/order_test.dart` — new `Order value equality` group: distinct instances from the same data are equal and share a `hashCode`, reflexivity, inequality against another type, one case per compared field (including `kind` and `fiatCode`), `copyWith` result equality, and collapsing in a `Set`. Written first and confirmed failing before the fix.
- `test/data/models/dispute_equality_test.dart` (new) — regression coverage for the second symptom: disputes holding equal-but-distinct orders now compare equal; different orders and order-vs-null stay unequal.
- `test/features/order/models/order_state_test.dart` — the pinned test flipped from `isNot(baseState())` to an equality assertion plus a `hashCode` comparison, with its comment updated.

## Rebuild-notification audit

Per the issue's "Watch out for" note, I reviewed every `state = ` in `lib/features/order/notifiers/` (`add_order_notifier.dart`, `order_notifier.dart`, `abstract_mostro_notifier.dart`). All of them assign through `updateWith`/`copyWith` on fields that participate in `OrderState.==`; none mutated something outside the compared set while relying on the identity-forced notification. No notifier changes were needed.

## Test plan

- [x] `flutter analyze` on the four touched files — no issues
- [x] `flutter test test/models/order_test.dart` — 11 passing (3 failing before the fix)
- [x] `flutter test test/data/models/dispute_equality_test.dart` — passing
- [x] `flutter test test/features/order/models/` — passing
- [x] Full `flutter test` — 960 passing, 1 pre-existing failure (see below)
- [x] CodeRabbit review on the diff — 0 findings
- [ ] CI green

### Pre-existing failure, unrelated to this PR

`test/notifiers/session_notifier_test.dart` fails to load locally because `test/mocks.mocks.dart` is stale and lacks `MockPushNotificationService`. It cannot be regenerated locally right now: `pub get` resolves `analyzer 8.4.1`, and `build_runner`'s build script fails to compile against it (`DartObjectImpl.getInvocation()` no longer exists). Verified with `git stash` that this failure reproduces on a clean tree at `main`, so it is out of scope here — CI regenerates mocks itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Order objects now compare by their values, including all order details.
  * Related order and dispute states now behave consistently when equivalent data is represented by separate objects.
  * Collections correctly recognize duplicate orders based on matching values.

### DIFF
--- a/lib/data/models/order.dart
+++ b/lib/data/models/order.dart
@@ -268,4 +268,48 @@ class Order implements Payload {
       createdAt: createdAt,
     );
   }
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    return other is Order &&
+        other.id == id &&
+        other.kind == kind &&
+        other.status == status &&
+        other.amount == amount &&
+        other.fiatCode == fiatCode &&
+        other.minAmount == minAmount &&
+        other.maxAmount == maxAmount &&
+        other.fiatAmount == fiatAmount &&
+        other.paymentMethod == paymentMethod &&
+        other.premium == premium &&
+        other.masterBuyerPubkey == masterBuyerPubkey &&
+        other.masterSellerPubkey == masterSellerPubkey &&
+        other.buyerTradePubkey == buyerTradePubkey &&
+        other.sellerTradePubkey == sellerTradePubkey &&
+        other.buyerInvoice == buyerInvoice &&
+        other.createdAt == createdAt &&
+        other.expiresAt == expiresAt;
+  }
+
+  @override
+  int get hashCode => Object.hashAll([
+        id,
+        kind,
+        status,
+        amount,
+        fiatCode,
+        minAmount,
+        maxAmount,
+        fiatAmount,
+        paymentMethod,
+        premium,
+        masterBuyerPubkey,
+        masterSellerPubkey,
+        buyerTradePubkey,
+        sellerTradePubkey,
+        buyerInvoice,
+        createdAt,
+        expiresAt,
+      ]);
 }
--- a/test/data/models/dispute_equality_test.dart
+++ b/test/data/models/dispute_equality_test.dart
@@ -1,0 +1,53 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mostro_mobile/data/models/dispute.dart';
+import 'package:mostro_mobile/data/models/enums/order_type.dart';
+import 'package:mostro_mobile/data/models/enums/status.dart';
+import 'package:mostro_mobile/data/models/order.dart';
+
+void main() {
+  Order order({Status status = Status.dispute}) => Order(
+        id: 'order-1',
+        kind: OrderType.sell,
+        status: status,
+        amount: 50000,
+        fiatCode: 'USD',
+        fiatAmount: 100,
+        paymentMethod: 'Wire transfer',
+        createdAt: 1700000000,
+        expiresAt: 1700003600,
+      );
+
+  Dispute dispute({Order? attached}) => Dispute(
+        disputeId: 'dispute-1',
+        orderId: 'order-1',
+        status: 'initiated',
+        order: attached,
+        adminPubkey: 'admin-pubkey',
+        adminTookAt: DateTime.utc(2026, 1, 1),
+        createdAt: DateTime.utc(2025, 12, 31),
+        action: 'dispute-initiated-by-you',
+      );
+
+  group('Dispute value equality with an attached order', () {
+    test('disputes holding equal-but-distinct orders are equal', () {
+      expect(dispute(attached: order()), equals(dispute(attached: order())));
+      expect(
+        dispute(attached: order()).hashCode,
+        equals(dispute(attached: order()).hashCode),
+      );
+    });
+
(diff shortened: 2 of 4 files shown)


grunch:
@coderabbitai review

PR #656, opened 2026-08-17
feat: show when order book filters are active

21Mill:
## Problem

The order book has five filters (fiat currency, payment method, premium/discount
range, reputation and account age), applied from the `OrderFilter` dialog. They
persist in memory for the whole session and stay applied when switching between
the BUY and SELL tabs, but the UI gave no hint that any of them were set: the
`FILTER` pill looked exactly the same with or without filters.

A user who forgot a filter was still applied saw fewer offers — or none at all —
and read it as missing liquidity on Mostro rather than as their own filtering.

## Solution

The pill now reflects the filter state:

- **No filters**: unchanged from today — grey pill, `FILTER │ N offers`.
- **Filters active**: Mostro green background, border, funnel icon (solid) and
  label, reading `FILTER (n)` with the number of active filters, plus an `✕`
  that clears all of them in a single tap without opening the dialog.

Filter pill: grey with no filters, green with FILTER (4) and a clear button

## Implementation notes

- `activeFilterCountProvider` counts filters whose state differs from their
  default, **reusing the exact predicates `filteredOrdersProvider` uses** to
  decide whether each filter block applies. That is the point that matters: if
  the two ever diverged, the pill would lie about what the list is showing.
- Those default values were duplicated in three places (provider declarations,
  the filtering guards, and the dialog's Clear button). They are now constants
  in `home_order_providers.dart`.
- The dialog's Clear button and the pill's `✕` share a single
  `clearAllOrderFilters` helper, so "no filters" has one definition.
- The `✕` lives outside the pill's main `InkWell` so tapping it resets the
  filters without also opening the dialog. It has a 40×40 touch target,
  a `Tooltip` and a `Semantics` label.
- The empty state (`noOrdersAvailable` / `tryChangingFilters`) is untouched.

## Localization

Adds `filterWithCount` and `clearFilters` to all six locales: en, es, it, fr,
de, pt.

## Testing

- New `test/features/home/home_order_providers_test.dart` — 14 tests covering
  each filter individually, range filters narrowed on only one bound, ranges
  restored to their defaults, several filters at once, and
  `clearAllOrderFilters`. All pass.
- `flutter analyze` — no new issues.
- Verified on a physical device (Nothing Phone 3a, Android): pill grey with no
  filters, green with `FILTER (4)`, and tapping `✕` cleared the filters, took
  the list from 2 to 3 offers and did **not** open the dialog.

## Summary by CodeRabbit

* **New Features**
  * Filter controls now show active filter counts and clearer visual states.
  * Added a convenient option to clear all order filters at once.
  * Filter actions are now accessible directly from the filter control.
* **Localization**
  * Added translated filter count and clear-filter labels in English, German, Spanish, French, Italian, and Portuguese.
* **Bug Fixes**
  * Resetting filters now consistently restores all default values.
  * Improved consistency when applying and clearing multiple filters.

### DIFF
--- a/lib/features/home/providers/home_order_providers.dart
+++ b/lib/features/home/providers/home_order_providers.dart
@@ -7,12 +7,63 @@ import 'package:mostro_mobile/shared/providers/order_repository_provider.dart';
 
 final homeOrderTypeProvider = StateProvider((ref) => OrderType.sell);
 
+// Default values for the order book filters. A filter is considered active
+// only when its state differs from these defaults.
+const double kDefaultRatingMin = 0.0;
+const double kDefaultRatingMax = 5.0;
+const double kDefaultPremiumMin = -10.0;
+const double kDefaultPremiumMax = 10.0;
+const int kDefaultMinDays = 0;
+
 // Filter state providers
 final currencyFilterProvider = StateProvider<List<String>>((ref) => []);
 final paymentMethodFilterProvider = StateProvider<List<String>>((ref) => []);
-final ratingFilterProvider = StateProvider<({double min, double max})>((ref) => (min: 0.0, max: 5.0));
-final premiumRangeFilterProvider = StateProvider<({double min, double max})>((ref) => (min: -10.0, max: 10.0));
-final minDaysFilterProvider = StateProvider<int>((ref) => 0);
+final ratingFilterProvider = StateProvider<({double min, double max})>(
+    (ref) => (min: kDefaultRatingMin, max: kDefaultRatingMax));
+final premiumRangeFilterProvider = StateProvider<({double min, double max})>(
+    (ref) => (min: kDefaultPremiumMin, max: kDefaultPremiumMax));
+final minDaysFilterProvider = StateProvider<int>((ref) => kDefaultMinDays);
+
+/// Number of filters that currently differ from their default value.
+///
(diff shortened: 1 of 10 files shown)


21Mill:
Applied both CodeRabbit nitpicks in 80e3602:

- The active-filter conditions now go through shared `_isMinDaysActive` / `_isRatingActive` / `_isPremiumActive` predicates, so `activeFilterCountProvider` and `filteredOrdersProvider` can no longer diverge.
- The premium and rating `RangeSlider`s take their bounds from the shared constants. The days `Slider` keeps its literals: its `20` has no constant and the value it receives is clamped, so the assertion risk does not apply there.

No behaviour change; `flutter analyze` clean and the filter provider tests pass.

21Mill:
Fixed in 768a858. The Clear button reset `minDays` to `kDefaultMinDays` but wrote `'0'` into the text field, so the two could disagree if the constant moved. Also applied the same change to the controller's initial text and to the fallback for unparseable input, which carried the same literal.

The days `Slider` keeps `min: 0.0, max: 20.0`: its `20` has no constant, so referencing one bound and hardcoding the other would not remove the divergence risk, and the value it receives is clamped.

Catrya (approved):
tACK

PR #659, opened 2026-08-19
fix: anchor transport resolution and authenticate v1 message senders

AndreaDiazCorreia:
## What

Makes the client's choice of wire transport depend only on evidence it can verify, and authenticates the sender of protocol v1 messages.

- Verify the signature of the node's kind-38385 info event before applying it, and ignore events that do not supersede the one already in use (NIP-01 addressable ordering, id tie-break included).
- Persist, per node pubkey, the last version that node was verified to assert *and* the asserting event's signed `created_at`, and resolve the transport from whichever assertion is the more recent evidence.
- Resolve the transport through a single entry point shared by the send path and the orders subscription, so the two cannot drift apart.
- Pin the seal author and verify its signature when unwrapping NIP-59 gift wraps — the same check mostro-core already performs daemon-side (`nip59.rs:201`).

## Behaviour change

A node whose `protocol_version` is unknown now resolves to v2 (kind 14) rather than v1, matching mostrod's own default since v0.18.0. Against a node genuinely running `transport = "gift-wrap"`, the client switches back to kind 1059 as soon as that node's signed info event arrives.

## The anchor is dated, not monotonic

A relay picks which signed events it serves, but it cannot mint one and cannot re-date one without breaking the signature. "Newer than anything this client has verified" is therefore exactly the evidence a relay cannot manufacture, and that — not the higher version number — is what the anchor compares:

- A replayed pre-migration v1 event loses against a recorded v2. This is the downgrade the PR exists to block.
- A **newer** signed v1 assertion wins. mostrod 0.18.x still ships `transport = "gift-wrap"`, so an operator undoing a bad v2 rollout publishes exactly that; a monotonic ratchet would refuse it and partition every client that had ever seen v2, silently, with no recovery short of reinstalling.
- Ties, and assertions that cannot be dated, fall back to the higher version.

Two things are never recorded: a version this client cannot resolve, and an absent tag. Resolution reads an absent tag as v1 because it can see the info event is in hand; the store outlives it, and persisting 1 would let a relay resolve v1 off remembered state alone by simply withholding the event.

Full rationale in `docs/architecture/TRANSPORT_V2_MIGRATION.md` §4.1.

## Notes

- `NostrService.decryptNIP59Event` and `NostrEvent.unWrap`/`mostroUnWrap` gained a required `expectedAuthor` argument. Run `dart run build_runner build -d` to refresh mocks.
- 17 atomic commits. `flutter analyze` clean; `flutter test` 1210 passing.

## Summary by CodeRabbit

* **Security**
  * Added sender authentication for encrypted messages, rejecting impostors and invalid signatures.
  * Validated incoming node information before accepting updates.

* **Reliability**
  * Preserved verified protocol versions across sessions and restarts.
  * Unknown or unsupported versions now safely use NIP-44 transport.
  * Improved handling of stale, duplicate, malformed, and legacy node information.
  * Improved peer and dispute chat delivery and message recovery.
  * Displays “Unknown” when a node’s protocol version is unavailable.

* **Tests**
  * Expanded coverage for authentication, signatures, transport selection, persistence, chat handling, and downgrade protection.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -125,15 +125,28 @@ The app follows a specific initialization order in `appInitializerProvider`:
 
 1. **NostrService Initialization**: Establishes WebSocket connections to configured relays
 2. **KeyManager Initialization**: Loads cryptographic keys from secure storage  
-3. **SessionNotifier Initialization**: Loads active trading sessions from Sembast database
-4. **SubscriptionManager Creation**: Registers session listeners with `fireImmediately: false`
-5. **Background Services Setup**: Configures notification and sync services
-6. **Order Notifier Initialization**: Creates individual order managers for active sessions
+3. **ProtocolVersionStore Initialization**: Loads the persisted transport anchor from SharedPreferences
+4. **SessionNotifier Initialization**: Loads active trading sessions from Sembast database
+5. **SubscriptionManager Creation**: Registers session listeners with `fireImmediately: false`
+6. **Background Services Setup**: Configures notification and sync services
+7. **Order Notifier Initialization**: Creates individual order managers for active sessions
 
 ### Critical Timing Requirements
 - SessionNotifier must complete initialization before SubscriptionManager setup
 - SubscriptionManager uses `fireImmediately: false` to prevent premature execution
+- `ProtocolVersionStore.init()` must complete before SubscriptionManager builds its first orders filter, since that filter's kind depends on the resolved transport
 - Proper sequence ensures orders appear consistently in UI across app restarts
+
+## Transport Resolution (Protocol v1/v2)
+
+Mostro nodes speak one of two wire transports, advertised in the kind-38385 info event via a `protocol_version` tag: v1 (NIP-59 gift wrap, kind 1059) or v2 (NIP-44 direct message, kind 14). Detection is automatic; there is no setting.
+
+- **Single resolution point**: every caller reads `anchoredProtocolVersionFor(ref)` (`features/mostro/protocol_version_store.dart`), never `MostroInstance.protocolVersion` directly. The send path and the receive subscription must not disagree about the transport — a client publishing kind 1059 while listening on kind 14 is partitioned from the node.
+- **Three tag states**: `NostrEvent.assertedProtocolVersion` reads a parseable value as that version, an absent tag on a *verified* event as v1 (a pre-0.18.0 daemon asserting by omission), and a present-but-unusable value as null. The last two are different facts and must not collapse.
+- **Unknown degrades upwards**: no info event, an unreadable value, or a version this client does not speak resolves to `kDefaultTransport` (v2), not v1. Any relay can produce "unknown" by withholding the info event, and v1's intake authenticates nothing, so resolving unknown to v1 would make omission a downgrade primitive.
+- **Persisted anchor**: `ProtocolVersionStore` remembers, per node pubkey, the last verified version *and* the asserting event's signed `created_at`. The more recent assertion wins, with ties and undatable assertions falling back to the higher version. A relay can replay a signed event but cannot re-date one, so a replayed pre-migration v1 event loses against a recorded v2 — while an operator genuinely rolling a node back to v1 is followed.
+- **Never recorded**: a version this client cannot resolve, and an absent tag. Persisting the latter would let a relay resolve v1 from remembered state alone by simply withholding the info event.
+- **Full detail**: `docs/architecture/TRANSPORT_V2_MIGRATION.md` §4.1, referenced from the code comments.
 
 ## Timeout Detection & Orphan Session Prevention
 
--- a/docs/architecture/TRANSPORT_V2_MIGRATION.md
+++ b/docs/architecture/TRANSPORT_V2_MIGRATION.md
@@ -69,7 +69,8 @@ manual override: detection is automatic (decision below).
 ### Decisions (fixed)
 
 1. **Transport selection: auto-detection only.** Read `protocol_version` from
-   the node's kind-`38385` event. `"2"` → v2; `"1"` or absent → v1. No settings
+   the node's kind-`38385` event. `"2"` → v2; `"1"` or an absent tag on a
+   verified event → v1; anything else → the safe default (§4.1). No settings
    toggle.
 2. **`version` field tied to transport.** Send `version: 1` on the gift-wrap
    path (preserves today's exact wire for already-deployed daemons) and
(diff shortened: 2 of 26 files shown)


AndreaDiazCorreia:
@codex review

grunch:
@AndreaDiazCorreia please rebase

grunch (changes requested):
## Review: fix: anchor transport resolution and authenticate v1 message senders

**Verification performed locally** (branch `pr-659` @ `14f83f26`, mocks regenerated): `flutter analyze` → 0 issues attributable to this PR (2 pre-existing `deprecated_member_use` infos in `test/core/automation/automation_contract_test.dart`); `flutter test` → **1114 passed, 0 failed**. Branch is 15 commits behind `main`, no overlapping files, merges cleanly — a rebase is still pending as requested.

The direction is right and the security reasoning in the doc comments is careful: seal-author pinning + signature verification closes a real forgery hole on the v1 path, and the info-event signature/replacement checks close the relay-forgery and rollback vectors. The findings below are about the *persistent* ratchet, which has one design gap I consider blocking, plus a few consistency items.

---

### 🟠 Major — `lib/features/mostro/protocol_version_store.dart` (whole file), `lib/features/mostro/transport.dart:73-95`

**The ratchet rejects evidence a relay cannot forge, and there is no recovery path.**

`anchoredProtocolVersion(advertised, remembered)` takes the max regardless of *when* each version was asserted. The stated threat model is "a relay replaying or suppressing events". But a kind-38385 event that is (a) signature-verified, (b) accepted by `_supersedesCurrentInfo`, and (c) has a `created_at` newer than anything this client has ever seen, is not something a relay can manufacture — it is the node itself asserting v1. Today that legitimate, newer, signed assertion is discarded.

Consequences:
- mostrod 0.18.x still ships `transport = "gift-wrap"`. Any operator who rolls back after a bad v2 rollout permanently partitions every client that ever saw v2 from their node. Symptom is silent: sends time out, orders subscription hears nothing.
- `ProtocolVersionStore.clear()` is never called from `lib/` (dead code). `removeCustomNode()` does not forget the node's entry, `selectNode()` does not either, and there is no settings action. The only way out is reinstalling the app.

Suggested fix, keeps the security property intact: persist `(version, createdAt)` per pubkey and only refuse a downgrade when the *advertising* event's `created_at` is **older than or equal to** the recorded one. A replayed pre-migration v1 event still loses (its timestamp predates the recorded v2), while a genuinely newer signed v1 event wins. Sketch:

```dart
int? anchoredProtocolVersion(AdvertisedVersion? advertised, RememberedVersion? remembered) {
  if (remembered == null) return advertised?.version;
  if (advertised == null) return remembered.version;
  // Newer signed evidence from the node itself is authoritative in either direction;
  // only an event that does not postdate the recorded one is a replay candidate.
  if (advertised.createdAt.isAfter(remembered.createdAt)) return advertised.version;
  return max(advertised.version, remembered.version);
}
```

At minimum, if you prefer to keep the strict monotonic design: wire `clear()` / a per-pubkey `forget(pubkey)` into `removeCustomNode()`, expose a reset in the node/about screen, and state the "operator rollback is unsupported" trade-off in `TRANSPORT_V2_MIGRATION.md`.

---

### 🟡 Minor — `docs/architecture/TRANSPORT_V2_MIGRATION.md:232-236`, `:276`, `:342-343`, `:367-369`

**Architecture doc now contradicts the code.** §4.1 still says absent/unreachable → `Transport.giftWrap (v1)` and "degrade to v1 on an unsupported value". This PR flips both to `kDefaultTransport = nip44` and adds a persisted ratchet, none of which is documented. `CLAUDE.md` also has no mention of `ProtocolVersionStore` or the new init-order requirement in `appInitializerProvider`. Please update both in this PR — the doc is referenced from the code comments (`§4.1`), so leaving it stale is actively misleading.

---

### 🟡 Minor — `lib/features/mostro/mostro_instance.dart:114`

**`MostroInstance.fromEvent` still collapses malformed → 1.** `protocolVersion: event.protocolVersion ?? 1` reads `protocol_version=abc` as v1, exactly the conflation this PR introduces `advertisesProtocolVersion` to avoid. Only the About screen consumes it today, so it is not a transport bug, but it will display "1" for a node whose transport was actually resolved to v2. Suggest making the model field nullable or applying the same three-state logic as `_advertisedBy` (absent → `kLegacyProtocolVersion`, unparseable → null).

---

### 🟡 Minor — `lib/features/mostro/transport.dart:45-55` + `lib/features/subscriptions/subscription_manager.dart:116-118`

**Unsupported versions are recorded into the ratchet.** `resolveTransport(3)` falls back to nip44 (fine), but `_recordAdvertisedProtocolVersion` will happily `record(pubkey, 3)`, after which `remembered == 3` forever and any later legitimate `protocol_version=2` is anchored to 3. Today the outcome happens to be the same transport, but the store now holds a value no resolver understands, and once a v3 with a different kind exists the client is pinned with no way back (same root cause as the Major finding). Consider only recording versions that `resolveTransport` maps to a concrete transport, i.e. `{1, 2}`.

---

### 🔵 Nit — `lib/data/repositories/open_orders_repository.dart:85-90`

`limit: 1` on the info filter is per-relay, so with N relays you may still receive N candidates; `_supersedesCurrentInfo` handles that correctly, but the comment above the filter reads as if a single event arrives — worth one line saying the tie-break also covers multi-relay delivery.

### 🔵 Nit — PR description

Says "7 atomic commits" and "1031 passing (+64)"; the branch has 12 commits and the suite is at 1114 on the current base. Please refresh after the rebase.

---

### Test coverage notes

- `nip59_authentication_test.dart`: the tampered-seal case now uses one ephemeral keypair for wrap author + encryption and asserts the exact `'Invalid seal signature'` message, so it genuinely reaches the seal check (earlier CodeRabbit finding resolved). 👍
- `open_orders_info_event_test.dart`: signature rejection, re-tagged forgery, NIP-01 tie-break and the two-filter / `since`-less info request are covered. 👍
- `protocol_version_store_test.dart`: reordering-safe write chain, load-merge, malformed persisted entries. 👍
- **Missing:** a test asserting what happens when a **newer signed v1 event** arrives after a recorded v2. That behaviour is the product decision at the heart of the Major finding and is currently untested; whichever direction you take, add the case explicitly.

---

**Verdict: Request changes** — for the ratchet design gap (Major) and the stale architecture doc; the rest are quick follow-ups. Happy to re-review right after the rebase.

Catrya (approved):
tACK

grunch:
protocolo v1 is deprecated we are not working on it anymore

PR #660, opened 2026-08-20
fix: authenticate events before recording them, and anchor message freshness

AndreaDiazCorreia:
## Depends on #659

This branch is cut from `fix/transport-downgrade-protection` and builds on the sender authentication landed there — the freshness work below is only meaningful once the signature covering `created_at` is actually verified. Merge #659 first, or review this against that branch.

## What

Removes the client's dependence on the relay for three things it was never entitled to decide: which events are unique, which are current, and which are genuinely signed.

- Record an event id in the dedup store only after the event is authenticated. Recording it first let a relay copy a genuine id onto a tampered event and claim the slot, so the real message behind it was dropped as a duplicate — free censorship, with nothing forged that had to survive a signature check.
- Take message freshness from the node-signed `created_at` (v2 kind 14) instead of local receive time, bound its age, and refuse any message older than the state it would modify. The high-water mark is persisted outside the databases a restore clears, so wiping history cannot make archived instructions look new.
- Verify signatures by recomputing the NIP-01 id at the remaining call sites, and fail closed on kind-0 node metadata.
- Never lower the trade key index from a daemon-supplied value: lowering it reissues trade keys that have already been used.
- Settle on one epoch unit for message timestamps.

## Notes

- Only the v2 path gains a signed clock. NIP-59 randomises the wrap and seal timestamps by design, so v1 keeps falling back to receive time.
- `decryptNIP59Event` no longer wraps authentication errors in a generic `Exception`, so "not from the node" is distinguishable from "malformed".
- One existing test asserted the unsafe behaviour (`applies unverified events with warning`) and was inverted.
- `flutter analyze` clean; `flutter test` 1111 passing.

### DIFF
--- a/lib/data/models/enums/storage_keys.dart
+++ b/lib/data/models/enums/storage_keys.dart
@@ -7,7 +7,8 @@ enum SharedPreferencesKeys {
   trustedNodeMetadata('trusted_node_metadata'),
   backgroundFilters('background_filters'),
   communitySelected('community_selected'),
-  nodeProtocolVersions('node_protocol_versions');
+  nodeProtocolVersions('node_protocol_versions'),
+  orderFreshness('order_freshness');
 
   final String value;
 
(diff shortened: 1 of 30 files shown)


AndreaDiazCorreia:
@coderabbitai review

grunch:
protocolo v1 is deprecated

PR #664, opened 2026-08-21
fix: check invoice amounts against the order's signed terms

AndreaDiazCorreia:
## What

The app never read the invoices it pays, and never read the amount it was asked to invoice for. Both figures came from the direct message that carried them, and both were acted on verbatim.

- **Seller side.** The pay screen showed one number and paid whatever the bolt11 encoded. A wallet honours the invoice, so the number on screen stated nothing about what would leave. Neither flow renders now unless the two agree — the guard sits on the screen, so the manual QR and copy paths are covered as well as the wallet one.
- **Buyer side.** The amount in the `add-invoice` message went straight into `make_invoice`, the Lightning address confirmation and the manual prompt. An invoice minted for it is what the trade settles at.

Both are now held against the terms the node signed rather than against the message: the order amount from the kind-38383 event, the fee rate from the kind-38385 one. mostrod computes both settlements from exactly those two inputs, so the client can re-derive them:

    fee            = round(feeRate × amount / 2)
    seller pays    = amount + fee
    buyer receives = amount − fee

Neither side is the order amount itself, which is why comparing against it would have refused every correct settlement.

## What the check is anchored to

Deriving from the signed events closes the gap between the message and the order, but both events are addressable: the node can republish either one after a trade is under way and move the figure the client accepts. Three further steps close what that leaves open.

**The terms are pinned at commitment.** `Session` records the order amount, the fee rate, and the three inputs the market quote is priced from — currency, fiat amount and premium — as they stood when it committed: at the take for a taker, at submit for a maker, where the maker's own figure is the agreement. All five live in addressable events, so republishing afterwards cannot move what the client will accept. Pricing the quote from the live event would have left the skim addressable in another direction: resolve fewer sats than the trade was agreed at, then republish the premium so the quote lands on the shaved figure and the gap disappears.

The one input still read live is a range order's fiat figure, which no commitment could have pinned — the band is not resolved to one figure until a taker settles it.

**The pins are durable before the commitment goes out.** Recording them in memory is not enough: the commitment is published as soon as the session is created, so a crash or a response timeout before the daemon replies would leave the trade standing remotely with its anchors gone. Restore is the harder case — it clears the session store outright, including sessions already written, and rebuilds each one from the node's own data, so a node A -> B -> A round trip returned every trade in flight to whatever terms the node currently advertises.

`SettlementTermsStore` is the part that survives both. It sits outside the session store, so clearing that does not take it, and it is keyed by trade key: the one identifier that spans the whole lifecycle, since it exists before the order id, it signs the commitment, and restore re-derives it from the key index. An anchor is written once — a retake must not move terms already agreed — and a record with nothing in it still counts, or a commitment made before the info event arrived would be indistinguishable from one that predates pinning.

**And the anchoring path fails closed.** "Durable before publish" only means anything if the failure stops the publish. A storage error that logged and returned would put a live remote trade behind anchors held only in memory, and any crash, timeout, restore or node switch would then bring it back reading whatever the node advertises — the exact state this is here to prevent. So the write propagates: the store drops what it could not persist rather than keeping a phantom that makes the next attempt a no-op, and the take, the create and the release that carries `NextTrade` all stop before publishing anything. The user is told the terms could not be saved and that nothing was sent.

Where failing closed would cost more than it protects, it does not. A range order's release carries the next leg's trade key, and that leg's anchor is a preparation, not the settlement: refusing to release funds already in escrow because the remainder's anchor could not be written trades the important thing for the incidental one. The remainder falls back to the terms a session predating pinning uses and the release goes out. A storage failure is not something the node can bring about, so this is not a downgrade it can reach for.

The session's own row stays best effort, deliberately. It is a fast path, not the guarantee: restore rebuilds the session from the node's data and takes its terms from the anchor, so losing the row costs a rebuild while losing the anchor costs the terms.

It also records *that* pinning ran, which is not the same as it having produced figures. A market-price order resolves no sats until after the take, and a commitment made before the info event arrives has no fee rate to pin, so an absent figure is not evidence that the client simply never looked. Where pinning ran and found nothing, the node had nothing to offer at the moment of agreement, and a value it publishes afterwards is not a term of it: the check reports itself unmade rather than adopting one. Sessions written before any of this read the marker as false and keep the behaviour they had.

**A rate the client was late to receive is not a rate supplied after the fact.** Refusing every absent fee rate put two unrelated things in one bucket. A node that withholds its kind-38385 and republishes it with a higher fee after the take has supplied a term after the agreement, which is what pinning exists to refuse. A node that published the same event, signed, well before the agreement, and whose client had not finished draining the order book, has supplied nothing: the client was behind. Against real nodes the second is what actually happens, and it left the caution standing on ordinary trades — a warning that fires on honest settlements is read once and furniture by the twentieth, which costs more than it buys the one time it fires for the reason it exists. It also fired more often the worse the connection, which is backwards for a warning about a node's conduct.

The event's own `created_at` separates them, measured against the instant the trade committed. A rate the node signed at or before that instant is adopted; one signed after it is not, and the caution stands. `created_at` is covered by the signature and by the recomputed id, so a node cannot move it without signing for it.

The instant of agreement comes from the durable anchor rather than from `Session.startTime`. Restore rebuilds every session with `DateTime.now()`, so that field records when the app last reconstructed the trade, not when the user agreed to it — a rate published after the agreement and before a restore would otherwise read as one that preceded it. A range order's remainder is anchored at its parent's instant rather than its own, so a rate signed between the parent's commitment and the release does not become a term of the child.

The info event is also asked for on its own subscription now, and first. It shared a filter with the order book, so it was answered somewhere inside a backlog of every order the node touched in 48 hours, which is how a client could commit before it arrived at all. Its own request drops `since` as well: the node republishes its info on its own schedule rather than the order book's, so one quiet for longer than the order window never announced itself.

**Market-price orders are re-priced against an outside rate.** A fixed-amount order states its sats up front, so the take screen shows them and the client pins them. A market-price order states only fiat and a premium, and the node resolves the sats itself after the take: the figure on the settlement screen is the first the user sees of it, and checking it against the node's own order event only asks the node whether it agrees with itself. Those orders are re-priced, mirroring mostrod's `get_market_quote` (premium included).

A gap wider than 5% refuses the flow, but only in the direction that runs against the user — a seller giving up more sats than the fiat is worth, a buyer receiving fewer. A gap in their favour is named and not stopped. The refusal keeps a way past it: the quote comes from a third party and can be stale or simply disagree, so the screen states the gap and leaves the decision with the user rather than making it for them.

The rate deliberately does not come from `exchangeServiceProvider`, which reads the node's own kind-30078 rates event first and only falls back to Yadio: a node that skims a settlement can publish the rate that makes the skim look correct. This asks Yadio directly.

The re-pricing applies only to sessions this build committed. One written by an earlier build pinned no sats either, so without that gate every one of them would fall through to the check — and a fixed-amount order among them, priced by hand and legitimately far from the market, would draw a caution on a settlement that is perfectly honest.

Continuing past a gap is consent to the figures the user saw, not to the screen. The approved quote is held — order id, settled sats, quoted sats — so a later republish or rate refresh that moves either figure asks again rather than riding on the earlier acknowledgement.

**A check that could not be made is said out loud.** Both screens previously fell through in silence when there was nothing to re-derive from. The node publishes both inputs and can withhold either, so that is a state it can put the screen into rather than only an unlucky race, and the absence of a refusal reads to the user as a confirmation. Both screens show a caution instead.

The market check reports four outcomes rather than one nullable figure: does not apply, still in flight, could not be made, checked. Collapsing them made three of them read to the user as a settlement that had passed — a rate still loading opened the flow before the check had an answer, and a currency tag the node had emptied looked the same as a clean result. Loading now holds the flow behind a progress state. "Could not be made" is a caution and does not refuse: an unreachable third party is not evidence against a settlement, and refusing on it would hand anyone who can break that request a way to stop honest trades.

The rate is fetched for the settlement being checked. Held for the process, the first rate ever fetched went on pricing every later trade, so a settlement hours later was compared against a stale price.

The payout screen is the exception, and deliberately: every check there cautions and none refuses. A settled order routes to it from the top of the add-invoice build, so it owns the retry after a failed payout and the window between release and payout. On both, the user has already given up their side, the hold invoice is settled, and collecting is all that remains — a screen that refused to mint would leave them unable to collect at all, with no path back. Naming the gap is the strongest thing that can be done there without stranding an honest trade.

Refusal and caution stay distinct throughout, and the line between them is what the check actually has against the settlement. A settlement that contradicts the signed terms is refused outright and offers only Cancel. A check that could not be completed only warns — an order with no amount to check the invoice against is an absence, not a disagreement, and refusing there would make it the stronger evidence of the two. An off-market gap refuses with a way through, except on the payout screen described above, where nothing refuses at all.

## Manual testing guide

**Setup.** Use a local mostrod and set `fee` to a non-zero value (e.g. `0.006`) in its settings. With `fee = 0.0` the seller amount, the buyer amount and the order amount are all equal, and none of the arithmetic above is exercised.

Every check here applies only to sessions created with this build, so test with fresh orders rather than trades that were already in flight. A session written by an earlier build carries no pinned terms and is deliberately left out: it goes on behaving exactly as it did, so a trade in flight when the user updates cannot draw a notice from a check that did not exist when it was agreed.

### Nothing should change on an honest node

This matters more than the attack cases: a spurious notice on every trade would be worse than the bug being fixed. None of these should show anything.

| Flow | Check |
|---|---|
| Make a sell order, have it taken, pay the hold invoice | Header and QR both show `amount + fee/2` |
| Make a buy order, have it taken, provide the invoice | Requested amount is `amount − fee/2` |
| Take a buy order (you are the seller) | Pay screen renders normally |
| Take a sell order (you are the buyer) | Add screen renders normally |
| Add-invoice via NWC, via Lightning address, and manually | All three start clean |
| Pay via NWC | Receipt shows the invoice's amount |
| Range order and market-price order | No notice, assuming the rate has not moved sharply |
| Range order taken in part, then settle the remainder | The child trade behaves like any other |
| Kill the app mid-trade and reopen straight onto the settlement screen | No notice should flash before the events arrive |
| Launch onto a cold cache and take or create straight away | No notice: a rate the node signed before the commitment is adopted even when it lands after it |

For the manual pay flow specifically: note the figure printed beside the QR code, scan it with a wallet, and confirm the wallet shows the same number. It previously printed the figure from the node's message.

### Security cases (patched mostrod)

All in `src/util.rs` unless noted.

**Seller-side overcharge — the original finding.** In `show_hold_invoice`, leave the message honest and inflate only the invoice, by creating the hold invoice with `new_amount * 10` instead of `new_amount`. Expected: the pay screen refuses, showing which two figures disagree, and offers only Cancel — no QR, no pay button.

**Buyer-side shave.** In `set_waiting_invoice_status`, set `buyer_final_amount` to 90% of its value. Expected: the add screen refuses.

**Republished order event.** In `show_hold_invoice`, halve `order.amount` before `new_amount` is computed, so the republished kind-38383, the message and the invoice all agree with each other. Without pinning this passed the check. Expected: the pay screen refuses, because the trade is held to what was pinned at the take.

**Fee rate changed mid-trade.** Take an order, then stop mostrod, raise `fee`, and restart it before paying. Expected: the trade proceeds with no refusal. Without pinning the expected amount moved and a correct invoice was rejected.

**Fee rate absent at commitment, signed afterwards.** Suppress the kind-38385 event, take an order, then let the node publish it with a raised `fee` and a fresh `created_at`. Expected: the settlement screen shows the "amount not verified" caution and keeps showing it. A rate signed after the commitment is not a term of it, so the caution does not clear when the event lands.

**Fee rate signed before the commitment, received after it.** Same setup, but republish the suppressed event unchanged, so its `created_at` still predates the take. Expected: no caution at all, and the settlement resolves normally. Nothing was supplied after the agreement; the client was behind. This is the case that fired on ordinary trades against honest nodes.

Repeat with a **market-price order**, where nothing is pinned on either side. The same must hold: an absent rate stays absent.

**Range remainder.** Create a range order while the kind-38385 is suppressed, have part of it taken and released, then let the node publish the event before the remainder settles. Expected: the child trade shows the caution too. It is anchored at its parent's instant of agreement, so a rate signed between that and the release is not a term of it either.

**Market skim.** Patch `get_market_quote` so the node resolves a figure away from the market, then create and take a market-price order.

| Patch | Seller (pay screen) | Buyer (add screen) |
|---|---|---|
| `Ok((sats as f64 * 0.8) as i64)` | Caution — the gap favours them | **Refuses** — fewer sats than the fiat is worth |
| `Ok((sats as f64 * 1.3) as i64)` | **Refuses** — more sats than the fiat is worth | Caution — the gap favours them |

On the refusing side, check that the pay button and QR are gone, that **Continue anyway** is offered beside Cancel, and that tapping it restores the flow while the notice stays visible in yellow.

Confirm the inverse too: a legitimate 10% premium must not trigger anything, since the formula accounts for it.

**Unreadable premium.** Publish a garbage `premium` tag on the kind-38383. Expected: no market notice at all and the trade proceeds — an unreadable premium means the order cannot be priced, not that it carries none. Pricing it as zero would misprice an honest trade by exactly the premium.

**Unavailable anchor.** Drop the `fee` tag from the kind-38385 entirely. Expected: a caution on both screens, no refusal; the pay screen still reconciles the invoice against the message's own figure.

**Unreadable and amountless invoices.** Return a malformed `payment_request`, and one with no amount. Expected: refusal in both cases, with different explanations under the same neutral title.

**Cancellation that fails.** Reach any refusal screen, then take the relays or mostrod down and tap Cancel. Expected: the screen stays, an error is shown, and the trade is still there to act on. It must not navigate home and leave the failure invisible.

### Environment

| Case | Expected |
|---|---|
| Airplane mode or `api.yadio.io` blocked, market-price order | No market notice, no spinner, no hang — the check is best-effort and never blocks on the rate |
| Switch Mostro instance between two nodes with different fees | The previous node's fee is not reused |
| Open the app onto a trade with slow relays | The fee resolves once the info event lands, unless the commitment predates it |
| Restart the app mid-trade | Pinned figures survive; the check behaves as it did before the restart |
| Retake an order after a failed payout | The original pin is not overwritten |

### Interface

- The six notice strings render in all six locales. German is the longest — check it does not overflow on a narrow screen.
- Cautions sit above the flow and leave it usable; refusals replace it.
- In dark mode the caution (yellow) and the refusal (red) are clearly distinguishable.
- Two cautions at once still lay out correctly.
- On the off-market refusal, Cancel and **Continue anyway** are both reachable and visually distinct.

## Not covered by this PR

- `pay_bond_invoice_screen` pays a bolt11 from a direct message with none of this applied. That is a separate finding.
- A market-price or range order pins no sats figure, since the node only resolves one after the commitment there was to make. Those orders rely on the market check, which refuses only the direction that runs against the user and can be overridden deliberately. Sessions committed by an earlier build are left out of it entirely.
- A range order's fiat figure is still read from the live event, because no commitment could have pinned it: the band is not resolved to one figure until a taker settles it. A node moving it there moves a term nobody had agreed to yet.
- Invoice expiry is not checked — the wallet enforces it — and the payee node is not surfaced, which overlaps with the invoice-swap finding.
- `Bolt11Network` is parsed but not enforced; pinning mainnet would break test instances.
- The fee-rate race is documented, not closed. mostrod computes `order.fee` at take time from its own settings while the client pins what the info event advertised, so an operator changing the fee against an unpropagated event makes the two disagree and refuses an honest trade. Accepting a second figure derived from the live rate would close it and is deliberately not done: that is the shape of a term supplied after the agreement, which pinning exists to refuse. The fix belongs on the daemon — publishing the fee that applied at take time alongside the order.
- A node willing to backdate its info event's `created_at` defeats the distinction between a rate it supplied after an agreement and one the client was late to receive. That case is warned about rather than blocked, which is what it was before; the trade is a warning that fires rarely and is therefore read, for one that fired on every trade and was not.
- Clock skew is not compensated. The two sides of that comparison come from different clocks, so a device running behind keeps the caution where an honest node published seconds before the commitment. That direction is the safe one, and the race being closed is a client draining a backlog, where the event predates the commitment by far more than any plausible skew.

## Notes

- No new dependency. `Bolt11Invoice` reads only the human-readable prefix, where the amount and network live; everything else sits in the bech32 data part behind a tagged-field encoding it deliberately leaves alone. Amounts are held in `BigInt` until a supply-cap check, since a long enough figure would otherwise wrap and land on a small plausible number.
- The rounding mirrors mostrod's, because a client that rounds differently would refuse correct settlements. The market quote likewise truncates where mostrod casts to `i64`.
- A missing signed input means the events have not both arrived, not that the request is wrong. The seller side falls back to reconciling the figure on screen against the invoice; the buyer side, which has no invoice yet, only refuses on an actual disagreement. Treating "cannot derive" as "wrong" would refuse correct payments on a slow relay.
- The NIP-47 pay request carries the invoice alone. The optional `amount` field is not handled uniformly — some wallets apply it only to a zero-amount invoice and error out when the invoice already encodes a figure — so sending it risks breaking honest payments to catch nothing: the invoice has already been reconciled against the signed terms before the request is built.
- The node's fee rate is followed rather than sampled, and pinned to the node currently selected, so a switch cannot leave the previous node's rate deriving amounts.
- Cancellation is awaited before the screen is left, so a failure is visible rather than thrown after disposal.
- The derived fee is bounded on the way out, not only on the way in. `feeRate.isFinite` on the input says nothing about the multiplication: an extreme finite rate overflows to infinity, where `round()` throws on a screen the user can reach, and one a little under that stays finite and saturates silently at the int64 ceiling — which then propagates into the derived amount as though the node had asked for it. `feeFor` returns null for both, distinct from a fee of zero, which is a real term a node can charge.
- Anchors in `SettlementTermsStore` are pruned after 90 days. A create stays deliberately ephemeral until the order is confirmed, as it was before; only its anchor is durable.
- A read that fails is treated differently from a value that cannot be parsed. An unread map may be perfectly good, so writing is refused rather than serializing memory over anchors belonging to trades still in flight — but the read is retried on the next pin rather than latched, since one transient failure at startup would otherwise leave the user unable to take, create or release for the rest of the run. An unparseable value has nothing behind it to preserve, and refusing to write would leave pinning broken for good, so it is overwritten.
- The order book request is bounded. Kind 38383 is addressable, so a relay holds one event per order rather than one per status change and the bound is not expected to be reached; it is there so a pathological book cannot make the first seconds after launch unbounded parsing on the main isolate.
- 19 new localization keys across all six locales.
- `flutter analyze` clean; `flutter test` 1337 passing.

## Summary by CodeRabbit

* **New Features**
  * Added invoice validation for unreadable, missing, or mismatched amounts.
  * Added market-rate checks with loading, unavailable, warning, refusal, and “Continue anyway” states.
  * Settlement calculations now honor terms confirmed when a trade session is committed.
  * Added Lightning invoice network and amount parsing.
  * Payments now use the amount encoded in valid invoices.

* **Bug Fixes**
  * Cancellation completes before navigation and reports failures clearly.

* **Localization**
  * Added related messages in English, German, Spanish, French, Italian, and Portuguese.


Catrya (changes requested):
Strong work. The pinning design is the right answer, and `termsPinned` as an explicit marker rather than something inferred from a null field is the non-obvious detail that makes it correct. The BOLT-11 prefix parser is well scoped, and the arithmetic genuinely mirrors mostrod where it matters — `f64::round` and Dart's `round()` are both half-away-from-zero, and the quote truncates like the `i64` cast. I checked both against the daemon source.

I'm requesting changes on two points, both small. On how I prioritised:

**No Mostro node is known to be dishonest today.** That means this PR's protective value right now is close to zero, while its regression surface is every honest trade that reaches either screen. As of today it can only cost us something. So I reviewed it primarily for "can this refuse or break a correct settlement?" rather than "does it catch an attacker?".

## Blocking

### 1. `amount` on every NIP-47 `pay_invoicbenefit

`lib/shared/widgets/nwc_payment_widget.dart

Every NWC payment request changes from `{in`. `amount` is optional in NIP-47 and wallet handling of it is not uniform: some impleme override for zero-amount invoices and return an error when the invoice already enusers have behaves that way, the NWC payment path breaks on every honest trade.

The benefit is nil. `InvoiceTerms.check` hale invoice, an amountless one, and one whose amount disagrees, before this line is reach the field protects against.

Please drop it, or gate it behind testing aub, Coinos, Zeus, Mutiny, Phoenix) before this lands. Removing it is a three-line cha in the PR.

### 2. The market check runs on sessions cr

`lib/features/order/providers/market_check_

```dart
if (session?.pinnedAmountSats != null) return null;
```

Every session written by an earlier build h, so it falls straight through into themarket check. A fixed-sats order can sit legitimately far from the market — the maker set the price by hand — so on the day this ships, trades already in flighal on a perfectly honest settlement.

The description acknowledges this obliquelyssions created with this build, so test with fresh orders"), but that is a QA instructior who is mid-trade when they update.

Gate it on `session?.termsPinned == true`. w protection, and declining to apply it to sessions that predate it is the conservativg the PR already applies in
`orderFeeRateProvider`.

## Not blocking, but worth deciding before
### 3. The unverified caution never clears, and may be more common than expected                                  By design, a commitment made before the kinPinned=true, pinnedFeeRate=null`, and thecaution stays for the life of the trade. Thpublished after the agreement is not a termof it.
But that couples the notice to how reliably the info event arrives at take time, and we have an open defect where subscriptions do not re-subscribe on a relawith any frequency, honest trades will carry a permanent yellow "Amount not verified" banner for a reason unrelated to settlement correctness. Worth measuring   before we ship a notice the user cannot dis
                                                                                                                    ### 4. Fee-rate race between the pin and mo
`order.fee` is computed at take time from `Settings::get_mostro().fee` (`take_sell.rs:178`, `take_buy.rs:132`,      `util.rs:81-86`), while the client pins whaertised. If an operator changes the fee andthe republished info event has not propagat, the two disagree — and the seller gets ahard refusal on an honest trade, recoverable only by cancelling.
                                                                                                               Narrow window, but pinning makes it permane the failure mode is the worst one in thePR. At minimum this deserves a comment. Better would be to consider whether a mismatch that stays within a plaufee band should caution rather than refuse.
                                                                                                               ### 5. The market check fails silent, and irice orders have
                                                                                                               For market-price and range orders `signedOr to the live published event(`settlement_anchor_provider.dart:46`), so e defence there. But `marketCheckProvider`returns null — producing neither a refusal  unreachable, when the currency tag isempty, or when the premium is unparseable. t, because `anchoredSats` is non-null inthat state.

That contradicts the principle the PR states for the fee path: "a check that could not be made is said out loud." Not urgent with no adversary today, but this is much cheaper to fix while the code is fresh than to rediscover
later. Suggest distinguishing "does not appand cautioning on the second.

### 6. `Role.seller` is hardcoded in the pa

`pay_lightning_invoice_screen.dart` uses `mr)` while the header two lines aboveconsults `session?.role`. Only sellers reach `/pay_invoice` today, so this is not a live bug, but the inconsistency
inside a single `build` invites a future onction inverts and an adverse gap silently
downgrades to a caution. Use the session roack.

### 7. `termsUnknown` is a hard refusal

On the pay screen `expectedSats = anchoredSats ?? orderState.order?.amount`; when both are absent the screeoutright. This is the one place where a misrather than a caution, against the rule the PR states elsewhere. Unlikely in practice, since the invoice and the amount ride the same DM, but inconsist### 8. The 5% tolerance and the post-failurThe normal add-invoice happens right after s. The re-prompt after a failed payout canarrive hours later, and BTC moves more thanisbehaving — a spurious refusal on the mostfragile path in the flow. There is an override, but consider widening the band or exempting the re-prompt. 
## Nits                                                                                                    
- `add_lightning_invoice_screen.dart:283-290` — the doc comment for `_buildBlockedFlow` ended up above     `_buildMarketBlockedFlow`, which now opens aphs; `_buildBlockedFlow` has none. Lookslike a rebase artifact.- Same file, lines 192-208 — indentation ofand its children is one level short.- Both screens duplicate roughly 30 near-ided` / `unverified` / `offMarket` /`marketBlocked` / `marketCaution`. A `Provi gate object would stop them drifting apart.
- `pinnedAmountSats` and `pinnedFeeRate` aression` (normalised in the constructor body) while `termsPinned` is `final`. For something called a pin, `late final` would match the intent.

## Verification performed

Checked out the head commit (`aa5819688f`)

| Check | Result |
|---|---|
| `flutter analyze` | Clean, apart from twotale `test/mocks.mocks.dart`(`MockPushNotificationService`), unrelated to this PR |
| New test suites | 100 tests pass. `market_check_provider_test` and `settlement_anchor_provider_test` fail to compile locally against the stale mocks (`clearPeerReputation`, from the taker-reputation merge), not because of
this PR — CI regenerates them |
| ARB parity for the 13 new keys | Complete
| `dart format` | Touched files do not passdoes the base commit — that is the
tall-style formatter, not this PR |
| Arithmetic against mostrod | Confirmed `g = order.amount + order.fee`,
`buyer_final_amount = order.amount.saturati`update_order_event` republishes the rawamount rather than `new_amount` — all match what the client derives |

## One alternative worth considering

Given that nothing is known to be attackinghip every refusal as a caution for a single
release. That keeps the whole detection val the notice appears and we hear about it —
with none of the regression risk, since no  If a cycle passes with no spurious noticesreported, we tighten to refusals with field evidence behind the decision instead of manual QA.

That deliberately blunts the PR, so it is aa review demand. But it fits the currentthreat model better than shipping enforceme

Catrya:
@AndreaDiazCorreia please fix the conficts

Catrya (changes requested):
I verified the central claim rather than taking it on trust. `isPayoutInvoice` is `paymentFailed || settledHoldInvoice`, and `_getStatusFromAction` maps an `addInvoice` carrying `settledHoldInvoice` to `Status.paymentFailed`, so the early return does fire on a real payout retry. The `isPayoutRetry` exemption in the add screen was unreachable, and the deleted test was indeed asserting on a state combination the flow does not produce. Relocating it to the payout screen as a caution is the right move, and cautions-only is the right call there: with nothing left to cancel, a refusal would leave the user unable to collect at all.

Both new keys are present across all six locales.

## Still blocking

Neither item from my last two reviews has changed. Restating them compactly.

### 1. Release and fiat-sent can fail silently

`lib/services/mostro_service.dart` — `releaseOrder` and `sendFiatSent` call `_prepareChildOrderIfNeeded` before publishing, unguarded. That chain reaches `createChildOrderSession` → `_anchorTerms` → `pin()`, which throws `SettlementTermsNotDurable` (`session_notifier.dart:460`). The call sites are fire-and-forget:

```dart
onPressed: () {
  Navigator.of(dialogContext).pop(true);
  ref.read(orderNotifierProvider(orderId).notifier).releaseOrder();
},
```

The user taps **Release**, the dialog closes, nothing is published, no message appears. The funds stay in escrow. Same shape at `trade_detail_screen.dart:402` for fiat-sent.

It needs a range order with a remainder and a SharedPreferences failure, so the trigger is rare — but it is silent, and it is the action that moves the money. Take and create already handle this exact throw with the `orderTermsNotStored` snackbar; these two paths were left out.

### 2. `_readFailed` latches for the whole process

`settlement_terms_store.dart` — `_readFailed` is set in `init()` when the initial read throws and is never retried, so `pin()` refuses for the rest of the run. One transient prefs error at startup leaves the user unable to take orders, unable to create them, and unable to release a range order until they restart the app.

### Suggested fix

Fail-closed is right for take and create: nothing has been published, and a retry costs nothing. For release and fiat-sent it is inverted — the trade is already live, and the child session is a *preparation* of the range order's next leg. Refusing to release because the child's anchor could not be persisted trades the important thing for the incidental one. The same reasoning `2e1a3c16` applies to the payout screen applies here: when there is no way back, naming the problem beats blocking the action.

- In `createChildOrderSession`, catch `SettlementTermsNotDurable` and leave the child with `termsPinned: false` — legacy behaviour — instead of aborting. The release goes out; the remainder falls back to the live events, exactly as a session predating pinning does.
- Have `pin()` attempt a re-read before refusing when `_readFailed` is set.

With those two, I approve.

## Verification

Checked out `2e1a3c16` locally.

| Check | Result |
|---|---|
| `flutter analyze` over `lib/` | 0 issues |
| Invoice screen, utils and settlement-store suites | All pass, including the new `payout_invoice_screen_test.dart` |
| ARB parity for `payoutAmountMismatchTitle` / `payoutAmountMismatchBody` | Complete across all six locales |
| `isPayoutInvoice` routing claim | Confirmed — the add screen's exemption was unreachable |
| Both blocking items | Confirmed unchanged at this head |

Catrya (changes requested):
Image

Tested this on a phone against 2 real Mostro nodes and I keep hitting the "Amount not verified" caution on ordinary trades. No patched daemon, nothing special set up, just taking an order. It shouldn't be showing up against honest nodes, and it's going to confuse people.

Also,  the new keys use voseo: "Revisalo vos antes de continuar", "Intentá de nuevo", "Revisá los términos antes de continuar". The rest of the app uses neutral tuteo

Here's what Fable found:

### What is actually happening

`anchoredSellerAmountProvider` needs two inputs, and returns null when either is missing. The one that goes missing is the fee rate: the kind-38385 info event had not arrived at the moment of the take, so the session is written with `termsPinned: true, pinnedFeeRate: null`, and `orderFeeRateProvider` reports the rate unknown for the life of the trade.

The reasoning for not clearing it is sound as far as it goes — a rate the node publishes *after* the agreement is not a term of it. But it puts two different situations in one bucket:

| | What happened |
|---|---|
| **Attack** | The node withheld the 38385 and published a new one, with a higher fee, after the take |
| **Race** | The event was already public and signed before the take; this client had not received it yet |

In the second case nothing was supplied after the fact. The fee was published and signed before the user committed. That the client had not finished downloading it is a network condition, not a term of the agreement.

### Why this is worse than a cosmetic annoyance

A caution that fires on honest trades against honest nodes teaches users to ignore it. By the twentieth trade it is furniture. The one time it fires for the reason it exists — a node withholding its fee event to move the figure — it will be dismissed unread, because it has never once meant anything before.

The notice cannot be dismissed and never clears, so it rides along for the whole trade.

### It gets worse the slower the connection

The window is from subscription open to the moment the user taps Take, and the info event shares that subscription with the order book:

```dart
NostrRequest(filters: [
  NostrFilter(kinds: [38383], since: now - 48h, authors: [...]),   // no limit
  NostrFilter(kinds: [38385], authors: [...], limit: 1),
])
```

The book filter is unbounded and asks for 48 hours. Since kind 38383 is addressable, that is not "the open orders" — it is every republish of every status change over two days. On a slow link the small event we need queues behind a lot of bytes we do not.

The deep-link path is the worst case: `mostro:` lands the user directly on `/take_sell/:orderId`, so they can commit seconds after launch with the subscription barely open.

So the notice fires **more often the worse the user's connection is**. For a P2P bitcoin app that is backwards: the users on the worst infrastructure are frequently the ones with the most at stake, and they are the ones who will see the yellow banner on every trade until it stops registering.

A security warning that correlates with bandwidth is not a security warning.

### Interaction with #663 — neither PR shows this on its own

Once #663 lands, every one of those kind-38383 events is signature-verified before admission, synchronously in the stream callback on the main isolate. I measured that verification at **16.3 ms per event** (100 signed 38383 events, after warmup, on an x86 desktop in the test VM — treat it as an order of magnitude, not a device number).

The info event therefore no longer queues only behind the bytes. It queues behind the CPU work of verifying everything that arrived ahead of it. A 300-event book burst is roughly 5 seconds of main-isolate work before the 38385 reaches `nodeFeeRateProvider`, and more on a mid-range phone.

**#663 widens exactly the window in which #664 shows this caution.** One makes intake slower, the other warns when intake was not fast enough. Neither author can see it from inside their own PR.

### What I would change

**1. Distinguish the race from the withholding, using `created_at`.** The kind-38385 carries a `created_at`, it is covered by the signature and the recomputed id (which #663 makes enforceable), and the session already records when it committed — `Session.startTime`, and `PinnedTerms.pinnedAt` in the new store.

```
info.createdAt <= session.startTime  ->  the rate was already public at commitment
                                         -> adopt it, no notice
info.createdAt >  session.startTime  ->  published afterwards
                                         -> keep the caution
```

Honest caveat: a malicious node can backdate `created_at`, so this does not hold against a determined attacker. But consider what is actually given up — that case is not blocked today either, only warned about. The trade is "a warning nobody reads because it always fires" for "a warning that fires rarely and is therefore read." That is the better deal, and it is the whole point of having the notice.

**2. Or wait for the info event before enabling Take**, with a short timeout. This moves the friction to before the commitment, where it costs nothing: the user has not agreed to anything yet. If the event still has not arrived when the timeout expires, then there genuinely are no terms and the caution is honest. This is arguably the cleaner fix, because the problem is a startup race and this addresses it at startup rather than warning about it for the life of the trade.

**3. Bound the order book filter with a `limit`.** 48 hours unbounded is a lot to ask of any connection, and after #663 it is also a lot of CPU. This one helps both PRs and is independent of the rest.

1 or 2 is what unblocks this from my side. 3 is a separate improvement worth having either way.

AndreaDiazCorreia:
Fixed all three, plus the voseo — four commits.

**Caution on honest trades** (`030d7b1e`). Took option 1. Two corrections that came out of it:

- Your screenshot is the **maker** path, not a take — that header only renders when `userIsMaker && justTaken` (`invoice_header.dart:26-30`). So the rate was missing at *submit*, which means option 2 would not have covered it: gating the Take button does not touch the create screen. Option 1 sits in `orderFeeRateProvider` and covers both entry points.
- `Session.startTime` cannot be the commitment instant. Restore rebuilds every session with `DateTime.now()` (`restore_manager.dart:680`), so after a restore it says when the app reconstructed the trade rather than when the user agreed — and a rate published after the agreement but before the restore would read as one that preceded it. It now comes from the durable anchor's `pinnedAt`, which restore does not touch. A range remainder is measured against its parent's instant rather than its own.

Backdating and clock skew are not compensated; both are named in the code and under "Not covered".

**Order book** (`a15c2278`, `98e7c8c5`). One correction: 38383 is addressable, so a relay following NIP-01 keeps only the latest event per `(pubkey, kind, d)` — the backlog is one event per order touched in 48h, not one per republish. Bounded it anyway, but the fix for the queueing was giving the info event its own request, sent first. Dropping `since` there also closed a latent bug: a node quiet for longer than the order window never announced itself at all.

**Voseo** (`914438df`). Six strings.

The two items from your previous review were already on the head you reviewed: `pin()` retries the read instead of latching `_readFailed`, and `createChildOrderSession` catches `SettlementTermsNotDurable` and falls back to `termsPinned: false` (`606e9b94`).

`flutter analyze` clean, 1337 tests passing. Description updated.

PR #665, opened 2026-08-21
fix: stop inbound payloads from restating what a trade is

AndreaDiazCorreia:
## Depends on #660

Cut from `fix/replay-and-censorship-protection`, which rewrote the two files this builds on. Merge #660 first, or review against that branch.

## What

Three payloads were taken at their word about things they should not have been able to state.

- **Trade terms.** An inbound Order payload replaced the tracked order outright, so any later message could restate every economic field — including the payment method the trade screen renders as the account to send fiat to. The fiat terms now freeze once the order leaves pending, since nothing in the protocol renegotiates them. The sats amount stays free: a market-price order has none until it is taken.

  A PaymentRequest's embedded order no longer becomes the tracked one either. Its amount is the figure for that payment — the hold invoice is the order plus the seller's fee, the payout the order less the buyer's — so it was never the trade. The pay screen now reads its figure from the payment request, where the node states it, rather than from an order the payment request had overwritten.

- **Bond amounts.** The node publishes in its kind-38385 event that it charges bonds, the percentage it takes and the floor it never goes below, and sizes every bond from exactly those. Nothing gated an inbound `pay-bond-invoice` on any of it. A request is refused when no policy was advertised, when it falls below the floor, and when it is not what the percentage yields for this order.

- **Trade key index.** Index 0 is what the master identity is derived from, so deriving "trade key 0" hands back that key and a session built on it signs and runs ECDH under the identity a trade depends on not sharing.  `setCurrentKeyIndex` has always refused it; derivation is reachable without going through the counter, and restore derives straight from an index in the response.

## Notes

- Where the node's arithmetic is mirrored, it is mirrored exactly — rounding included — because a client that computes differently refuses figures that are correct.
- The bond amount check needs the order's sats amount. A market-priced range order is sized from the taker's own quote, which the node computes and never sends, so only the advertised floor is checked there.
- The bond is held against the amount the node declares, not against the decoded invoice. Binding the two needs the BOLT-11 parser from the invoice PR; it is a small follow-up once that lands.
- Conflicts with the invoice PR on `pay_lightning_invoice_screen.dart`: both change where the payment figure comes from. Keep this branch's reading from `paymentRequest.order`.
- 11 existing test fixtures used trade index 0 as filler; none asserted it was valid.
- `flutter analyze` clean; `flutter test` 1215 passing.

### DIFF
--- a/lib/data/models/order.dart
+++ b/lib/data/models/order.dart
@@ -20,6 +20,7 @@ class Order implements Payload {
   final String? buyerTradePubkey;
   final String? sellerTradePubkey;
   final String? buyerInvoice;
+
   /// Seconds since the Unix epoch, as the protocol sends it (Nostr
   /// convention). Convert before handing it to anything that expects
   /// milliseconds, including [MostroMessage.timestamp].
@@ -166,9 +167,8 @@ class Order implements Payload {
       return Order(
         id: parseOptionalStringField('id'),
         kind: OrderType.fromString(parseStringField('kind')),
-        status: statusRaw != null
-            ? Status.fromString(statusRaw)
-            : Status.pending,
+        status:
+            statusRaw != null ? Status.fromString(statusRaw) : Status.pending,
         amount: amount,
         fiatCode: parseStringField('fiat_code'),
         minAmount: minAmount,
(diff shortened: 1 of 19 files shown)


grunch:
protocolo v1 is deprecated

PR #667, opened 2026-08-21
feat: dedicated payout invoice flow for settled orders

Catrya:
## Problem

Once Mostro settles the hold invoice, the buyer may still need to supply a new
Lightning invoice: either the payout failed after exhausting its retries, or
they realise the invoice/address they gave is not going to work. The app reused
the add-invoice screen from the take handoff for that, and the mismatch shows:

- It announced "A seller has taken your buy order" and displayed the
  counterpart's reputation, neither of which has anything to do with collecting
  a payout. Worse, it did so only for the maker, so the same moment looked
  different depending on the role.
- It offered a Cancel button that Mostro rejects with `NotAllowedByStatus` once
  the order is settled (`cancel.rs`), so the user got a `cant-do` and nothing
  else.
- It re-offered the Lightning address shortcut, which may be exactly what broke
  the payout.
- And in most of those states the order detail showed no button at all, so
  there was no way into the screen to begin with.

On top of that, the retry was detected from the previous `payment-failed`
notice, which the daemon only sends on the *first* failure. A closed, restored
or reinstalled app never saw it and treated the retry as a fresh take.

## Changes

1. **`fix(trades): don't block order details waiting for the public event`** —
   the screen required the public 38383 event, which the order book only caches
   for 48h, but uses it solely for the creator reputation card of a pending
   order. A settled, canceled or old order left it stuck on a spinner with no
   app bar and no way out. Prerequisite here, since the new flow navigates to
   the order detail after submitting.
2. **`fix(order): detect the payout add-invoice from the message payload`** —
   the daemon only sends this action with a `settled-hold-invoice` payload when
   it is asking for a payout invoice, so the message identifies the case on its
   own.
3. **`feat(order): dedicated screen to collect the sats of a settled order`** —
   new screen: neutral wording, no counterpart context, no Lightning address
   shortcut, nothing to cancel. Reached both after a failed payout and while
   the payment is still on its way, so it never claims that something failed.
   The failure notice stays on the order detail, where it is conditional and
   already translated.
4. **`feat(trades): keep the invoice available while the order is settled`** —
   FSM rows for `released` and `payment-failed`, which offered no action at
   all, and `cancel` dropped from `settled-hold-invoice` for both roles.

## Not included

A contextual message for the `cant-do` Mostro returns when an invoice is submitted while a payout attempt is in flight. The window is short and self-clearing (the marker is released on both success and failure), so the worst case is a generic error and a retry. Worth its own PR, together with the other context-free `cant-do` messages.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated payout invoice screen for settled orders.
  - Supports automatic invoice generation or manual invoice entry.
  - Added “Collect Your Sats” guidance and payout details.
  - Cancel controls now appear only when cancellation is available.

- **Bug Fixes**
  - Improved handling of payment-failed and payout-retry states.
  - Buyers can retry or add invoices across supported payout statuses.
  - Sellers can no longer cancel settled hold-invoice actions.

- **Localization**
  - Added payout and sats-collection translations in English, German, Spanish, French, Italian, and Portuguese.

### DIFF
--- a/lib/features/order/models/order_state.dart
+++ b/lib/features/order/models/order_state.dart
@@ -322,9 +322,10 @@ class OrderState {
         return Status.waitingBuyerInvoice;
       
       case Action.addInvoice:
-        // If current status is paymentFailed, maintain it for UI consistency
-        // Otherwise, transition to waitingBuyerInvoice for normal flow
-        if (status == Status.paymentFailed) {
+        // Mostro reuses this action to ask for a payout invoice after a failed
+        // payment, and only then does the payload carry settled-hold-invoice.
+        if (payloadStatus == Status.settledHoldInvoice ||
+            status == Status.paymentFailed) {
           return Status.paymentFailed;
         }
         return Status.waitingBuyerInvoice;
--- a/lib/features/order/models/order_state.dart
+++ b/lib/features/order/models/order_state.dart
@@ -621,6 +621,12 @@ class OrderState {
         Action.paymentFailed: [
           Action.addInvoice,
         ],
+        // Mostro acked the new invoice (invoice-updated, no payload) and the
+        // payout is being retried. The swap is accepted between attempts, so
+        // keep the way in for a buyer whose second invoice is wrong too.
+        Action.invoiceUpdated: [
+          Action.addInvoice,
+        ],
       },
       Status.active: {
         Action.holdInvoicePaymentAccepted: [
@@ -721,6 +727,11 @@ class OrderState {
         // Released but not paid yet: let the buyer replace an invoice that is
         // not going to work instead of waiting for the payout to fail.
         Action.released: [
+          Action.addInvoice,
+        ],
+        // Same as the payment-failed row: the invoice-updated ack must not
+        // leave the buyer without a way back in.
+        Action.invoiceUpdated: [
           Action.addInvoice,
         ],
       },
--- a/lib/features/order/models/order_state.dart
+++ b/lib/features/order/models/order_state.dart
@@ -577,7 +577,6 @@ class OrderState {
       Status.settledHoldInvoice: {
         Action.addInvoice: [
           Action.addInvoice,
-          Action.cancel,
         ],
       },
     },
@@ -614,11 +613,14 @@ class OrderState {
         ],
       },
       Status.paymentFailed: {
+        // Only the invoice: the order is already settled, so Mostro rejects a
+        // cancel and there is nothing left to dispute.
         Action.addInvoice: [
-          // Only allow add invoice, no cancel or dispute during retrying
           Action.addInvoice,
         ],
-        Action.paymentFailed: [],
+        Action.paymentFailed: [
+          Action.addInvoice,
+        ],
       },
       Status.active: {
         Action.holdInvoicePaymentAccepted: [
@@ -715,7 +717,11 @@ class OrderState {
       Status.settledHoldInvoice: {
         Action.addInvoice: [
           Action.addInvoice,
-          Action.cancel,
+        ],
+        // Released but not paid yet: let the buyer replace an invoice that is
+        // not going to work instead of waiting for the payout to fail.
+        Action.released: [
+          Action.addInvoice,
         ],
       },
     },
--- a/lib/features/order/screens/add_lightning_invoice_screen.dart
+++ b/lib/features/order/screens/add_lightning_invoice_screen.dart
@@ -62,7 +62,9 @@ class _AddLightningInvoiceScreenState
         if (isPayoutInvoice(orderState.status)) {
           return PayoutInvoiceScreen(
             orderId: orderId,
-            order: orderPayload,
+            // The stream only matches messages whose payload is an Order; the
+            // notifier also keeps one carried by a PaymentRequest.
+            order: orderPayload ?? orderState.order,
           );
         }
         final amount = orderPayload?.amount;
--- a/lib/features/order/screens/add_lightning_invoice_screen.dart
+++ b/lib/features/order/screens/add_lightning_invoice_screen.dart
@@ -58,8 +58,9 @@ class _AddLightningInvoiceScreenState
         final orderPayload = mostroMessage?.getPayload<Order>();
         final orderState = ref.watch(orderNotifierProvider(orderId));
         // A settled order is not asking for the invoice a take requires, but
-        // for the payout one, which has its own screen.
-        if (isPayoutInvoice(orderState.status)) {
+        // for the payout one, which has its own screen. It deliberately ignores
+        // `lnAddress`: the address may be what broke the payout to begin with.
+        if (orderState.status.isPayoutInvoice) {
           return PayoutInvoiceScreen(
             orderId: orderId,
             // The stream only matches messages whose payload is an Order; the
--- a/lib/features/order/screens/add_lightning_invoice_screen.dart
+++ b/lib/features/order/screens/add_lightning_invoice_screen.dart
@@ -3,6 +3,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
 import 'package:mostro_mobile/core/app_theme.dart';
 import 'package:mostro_mobile/features/order/providers/order_notifier_provider.dart';
+import 'package:mostro_mobile/features/order/screens/payout_invoice_screen.dart';
 import 'package:mostro_mobile/features/order/widgets/order_app_bar.dart';
 import 'package:mostro_mobile/features/wallet/providers/nwc_provider.dart';
 import 'package:mostro_mobile/shared/providers/mostro_storage_provider.dart';
@@ -55,6 +56,15 @@ class _AddLightningInvoiceScreenState
     return mostroOrderAsync.when(
       data: (mostroMessage) {
         final orderPayload = mostroMessage?.getPayload<Order>();
+        final orderState = ref.watch(orderNotifierProvider(orderId));
+        // A settled order is not asking for the invoice a take requires, but
+        // for the payout one, which has its own screen.
+        if (isPayoutInvoice(orderState.status)) {
+          return PayoutInvoiceScreen(
+            orderId: orderId,
+            order: orderPayload,
+          );
+        }
         final amount = orderPayload?.amount;
         final fiatAmount = orderPayload?.fiatAmount.toString() ?? '0';
         final fiatCode = orderPayload?.fiatCode ?? '';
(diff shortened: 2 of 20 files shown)


lib/features/order/models/order_state.dart
```diff
@@ -613,11 +613,14 @@ class OrderState {
...
       },
       Status.paymentFailed: {
+        // Only the invoice: the order is already settled, so Mostro rejects a
+        // cancel and there is nothing left to dispute.
         Action.addInvoice: [
-          // Only allow add invoice, no cancel or dispute during retrying
           Action.addInvoice,
         ],
-        Action.paymentFailed: [],
+        Action.paymentFailed: [
```
grunch, line 621: 🔴 **Blocking — the button vanishes right after the buyer submits the invoice.**

When the buyer sends the new invoice, mostrod persists it via `pay_new_invoice` and replies with `Action::InvoiceUpdated` and **no payload** (`add_invoice.rs:47-55`). On our side `invoiceUpdated` keeps the status (`_getStatusFromAction`, `payloadStatus ?? status`) but the action becomes `invoiceUpdated`, and there is no `Action.invoiceUpdated` row here nor under `Status.settledHoldInvoice` → `getActions()` returns `[]` and **COLLECT SATS** disappears from the order detail until the next `payment-failed` (if the retry fails again) or `purchase-completed`.

That is exactly the "released but not paid yet" window this PR wants to cover: a buyer who realises the *second* invoice is wrong too has no way back in.

```suggestion
        Action.paymentFailed: [
          Action.addInvoice,
        ],
        // Mostro acked the new invoice (invoice-updated, no payload) and the
        // payout is being retried: keep the way in until it succeeds or fails.
        Action.invoiceUpdated: [
          Action.addInvoice,
        ],
```

The same row is needed under `Status.settledHoldInvoice` for the buyer (line 723 block). Please add a test to `order_state_settled_actions_test.dart` for the sequence `released → addInvoice (settled payload) → invoiceUpdated` still offering `Action.addInvoice`.

lib/features/order/models/order_state.dart
```diff
@@ -714,7 +717,11 @@ class OrderState {
       Status.settledHoldInvoice: {
         Action.addInvoice: [
           Action.addInvoice,
-          Action.cancel,
+        ],
+        // Released but not paid yet: let the buyer replace an invoice that is
+        // not going to work instead of waiting for the payout to fail.
+        Action.released: [
```
grunch, line 729: ⚠️ Two things on this row:

1. Same gap as above: once the buyer submits and mostrod answers `invoice-updated`, this state becomes `settledHoldInvoice / invoiceUpdated`, which has no row → no button. An `Action.invoiceUpdated: [Action.addInvoice]` row is needed here too.

2. In practice `release` dispatches `do_payment` immediately and `pay_new_invoice` is guarded by `payout_payment_hash IS NULL` (`add_invoice.rs:28-44`), so the button offered in the `released` window will almost always yield a context-free `cant-do`. The PR acknowledges this under "Not included", but offering an action that nearly always fails with a generic error is arguably worse UX than not offering it. Consider gating this row behind the contextual `cant-do` message, or state the accepted trade-off explicitly in the PR description.

lib/features/order/models/order_state.dart
```diff
@@ -576,7 +577,6 @@ class OrderState {
       Status.settledHoldInvoice: {
```
grunch, line 577: 🧹 After this PR the `Status.settledHoldInvoice → Action.addInvoice` rows (this one for the seller, and the buyer one at line 717) are unreachable: any `add-invoice` whose payload is settled now maps to `Status.paymentFailed`, and `OrderNotifier.sync()` always rebuilds from the initial state through `updateWith` (`OrderState.fromMostroMessage` has no callers in `lib/`). The seller never adds an invoice anyway. Either drop them or leave a comment explaining why they are kept.

lib/features/order/screens/payout_invoice_screen.dart
```diff
@@ -0,0 +1,189 @@
...
+import 'package:mostro_mobile/shared/widgets/nwc_invoice_widget.dart';
+
+/// Whether this add-invoice asks for the payout of an already settled order
+/// rather than the invoice a take requires.
+///
+/// Both statuses mean the same thing: the hold invoice is settled and the only
+/// thing left is paying the buyer. `paymentFailed` is where a payout retry
+/// lands, `settledHoldInvoice` is the window between the release and the
+/// payout, where the buyer may want to replace a wrong invoice on their own.
+bool isPayoutInvoice(Status status) =>
```
grunch: 🏗️ `isPayoutInvoice` is a domain predicate over `Status`, but it lives in a screen file and is imported by `trade_detail_screen.dart` and `add_lightning_invoice_screen.dart`. The same rule is already encoded independently in `_getStatusFromAction` and in `notification_data_extractor.dart:46-47` (`order?.status == Status.settledHoldInvoice`).

Suggest moving it next to the state it describes, e.g. `bool get isPayoutInvoice` on `OrderState` (or an extension on `Status`), and reusing it from the three places so the rule has a single owner.

lib/features/order/screens/payout_invoice_screen.dart
```diff
@@ -0,0 +1,189 @@
...
+            ? Column(
+                crossAxisAlignment: CrossAxisAlignment.start,
+                children: [
+                  header,
+                  const SizedBox(height: 24),
+                  NwcInvoiceWidget(
+                    sats: sats,
+                    orderId: orderIdValue,
+                    onInvoiceConfirmed: (invoice) async {
+                      await _submitInvoice(invoice, sats);
```
grunch: 🔧 Inconsistent amount between the two submit paths: the NWC path sends `sats` (which falls back to `0`), while the manual path at line 113 sends `order?.amount` (nullable, per 7cefdff9). Harmless today because `showNwcInvoice` requires `sats > 0`, but the fix in 7cefdff9 only covered one of the two paths — please unify on `order?.amount`.

lib/features/order/screens/add_lightning_invoice_screen.dart
```diff
@@ -55,17 +56,26 @@ class _AddLightningInvoiceScreenState
     return mostroOrderAsync.when(
       data: (mostroMessage) {
         final orderPayload = mostroMessage?.getPayload<Order>();
+        final orderState = ref.watch(orderNotifierProvider(orderId));
+        // A settled order is not asking for the invoice a take requires, but
+        // for the payout one, which has its own screen.
+        if (isPayoutInvoice(orderState.status)) {
```
grunch: 📝 In payout mode `widget.lnAddress` (from `/add_invoice/:orderId?lnAddress=`) is silently ignored. That is intentional per the PR description (the LN address may be what broke the payout), but a one-line comment here would stop someone from "fixing" it later.

test/shared/widgets/add_lightning_invoice_widget_test.dart
```diff
@@ -0,0 +1,66 @@
...
+
+    test('is false for the invoice a take asks for', () {
+      expect(isPayoutInvoice(Status.waitingBuyerInvoice), isFalse);
+      expect(isPayoutInvoice(Status.waitingPayment), isFalse);
+      expect(isPayoutInvoice(Status.active), isFalse);
+      expect(isPayoutInvoice(Status.pending), isFalse);
+    });
+  });
+
+  group('AddLightningInvoiceWidget cancel button', () {
```
grunch, line 8: 🧪 This file is named after `PayoutInvoiceScreen` but never pumps it — it only covers `isPayoutInvoice` and the shared widget's cancel button. Please add at least one widget test of the screen itself (header rendered from the `order` snapshot, no cancel button, `/trade_detail/<id>` navigation after a successful submit).

Also uncovered by the PR:
- `AddLightningInvoiceScreen` branching into `PayoutInvoiceScreen`, including the `orderPayload ?? orderState.order` fallback introduced in 7cefdff9.
- `TradeDetailScreen`: the `collectSatsButton` label switch, and the fact that it no longer blocks on `eventProvider == null` (commit 9aebae00 is a behaviour fix on its own and deserves a regression test).

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -382,8 +385,12 @@ class TradeDetailScreen extends ConsumerWidget {
 
         case actions.Action.addInvoice:
           if (userRole == Role.buyer) {
+            // On a settled order the invoice is not part of the trade any
+            // more: it is the only way left to collect, so the label says so.
             widgets.add(_buildNostrButton(
-              S.of(context)!.addInvoiceButton,
+              isPayoutInvoice(tradeState.status)
```
grunch: 📝 Related, outside the diff: `case actions.Action.paymentFailed` at lines 583-586 still says "No additional buttons (Add Invoice, Cancel, Dispute) should appear", which now contradicts the new `paymentFailed → [addInvoice]` FSM row. That `case` is also dead code (the switch iterates `getActions()`, which never returns `paymentFailed`), so it can be removed together with the stale comment.

grunch (changes requested):
## Review summary

Reviewed locally against `main` (merge-base `c3c2d7a`), with the daemon behaviour cross-checked in mostrod (`src/app/release.rs`, `src/app/add_invoice.rs`, `src/app/cancel.rs`).

**Verified**
- The payout-retry `add-invoice` really carries `Payload::Order` with status `settled-hold-invoice` and a fee-net `amount` (`release.rs:136-146`), so detecting it from the payload is correct and covers the closed/restored/reinstalled app case. The three tests in `order_state_payout_retry_test.dart` are the right ones.
- `cancel` on a settled order is rejected with `NotAllowedByStatus` (`cancel.rs`), so dropping it from the FSM rows is correct.
- The buyer receives `released` with no payload, so the new `released → addInvoice` row is the right hook for that state.
- `flutter analyze` clean on the touched files; the PR's test directories pass (209/209). The only failures in the full suite are stale-mock compile errors in `test/notifiers/*` unrelated to this PR.
- All six ARBs carry the three new keys, placeholder naming matches the existing `snake_case` convention, `gen-l10n` reports no untranslated messages.

**Blocking**
- 🔴 The new flow breaks right after the buyer submits the invoice: mostrod answers with `invoice-updated` (no payload), and neither `Status.paymentFailed` nor `Status.settledHoldInvoice` has an `Action.invoiceUpdated` row, so `getActions()` returns `[]` and the **COLLECT SATS** button disappears until the next `payment-failed` / `purchase-completed`. See inline comment.

**Should fix**
- `released` window: `pay_new_invoice` requires `payout_payment_hash IS NULL`, and `release` dispatches `do_payment` immediately, so the button offered in that window will almost always produce a context-free `cant-do`. Worth either holding that row until the contextual message exists or spelling out the trade-off in the PR.
- Test coverage gaps (inline): `PayoutInvoiceScreen` itself, the `AddLightningInvoiceScreen` branching, and the two behaviour changes in `TradeDetailScreen` (label switch, no longer gated on the 38383 event) have no tests.
- `isPayoutInvoice` is a domain predicate living in a screen file and imported by two other screens; `notification_data_extractor.dart:46-47` already encodes the same rule independently.

**Nits / docs (not in the diff, so noted here)**
- `lib/features/trades/screens/trade_detail_screen.dart:583-586`: the comment on `case Action.paymentFailed` ("No additional buttons (Add Invoice, Cancel, Dispute) should appear") now contradicts the FSM. That `case` is also dead code: the switch iterates over `getActions()`, which never yields `paymentFailed`.
- `docs/architecture/ORDER_STATUS_HANDLING.md:58,100` still describe the previous-status detection ("unless in paymentFailed state"); should describe the payload-based rule.
- `docs/automation-contract.md:83` lists `invoice.cancel` as a fixed part of invoice entry; it is now conditional (absent in the payout flow).

Issue #606, opened 2026-06-01
Dart run build_runner build fails on Flutter 3.44.0, source_gen 3.1.0 incompatible with analyzer 8.x


codaMW:
### Environment

- Flutter: 3.44.0 (channel stable)
- Dart SDK: 3.12.0
- OS: Ubuntu 24.04 (WSL2)

### Steps to Reproduce

```bash
git clone https://github.com/MostroP2P/mobile.git
cd mobile
flutter pub get
dart run build_runner build -d
```

### Error

```
W source_gen-3.1.0/lib/src/constants/revive.dart:82:40: Error: The method
  'getInvocation' isn't defined for the type 'DartObjectImpl'.
   - 'DartObjectImpl' is from 'package:analyzer/src/dart/constant/value.dart'
  Try correcting the name to the name of an existing method, or defining a
  method named 'getInvocation'.
    final i = (object as DartObjectImpl).getInvocation();
                                         ^^^^^^^^^^^^^
E Failed to compile build script.
```

### Root Cause

`build_runner: ^2.4.0` resolves to `build_runner 2.7.1` which pulls `source_gen 3.1.0`. That version calls `DartObjectImpl.getInvocation()` which was removed in `analyzer 8.x`. Flutter 3.44.0 ships Dart 3.12.0 which forces `analyzer 8.4.1`.

`source_gen 3.2.0` does not exist on pub.dev, there is no released version of `source_gen` that supports `analyzer >=8.0.0`. Bumping `build_runner` or `mockito` constraints alone does not resolve this because the underlying `source_gen` package has not yet published a compatible release.

This was verified on the latest commit of `main` (`e46854f3`). Android Studio is not involved, this is pure Dart code generation and fails on a clean clone.

### Impact

`dart run build_runner build -d` is a required setup step per the README. This blocks all contributors on Flutter stable 3.44.0 from building the app locally.

### Possible Fix

Once `source_gen` publishes a release supporting `analyzer 8.x`, bumping `build_runner` and `mockito` to their latest versions should resolve this. Alternatively, replacing `mockito` with a code generation approach that does not depend on `source_gen 3.x` would also fix it. This is related to the broader dependency modernisation tracked in #512.

PR #668, opened 2026-08-21
fix: never let marking a dispute read block opening it

### LINKED ISSUE #653: DisputeListItem: a failing read-status write silently blocks opening the dispute
## Context

Found while writing widget tests for the dispute list (PR #651).

`lib/features/disputes/widgets/dispute_list_item.dart`:

```dart
return GestureDetector(
  onTap: () async {
    // Mark dispute as read when user opens it
    await DisputeReadStatusService.markDisputeAsRead(dispute.disputeId);
    onTap();
  },
  ...
```

`DisputeReadStatusService.markDisputeAsRead` (`lib/services/dispute_read_status_service.dart`) does:

```dart
final prefs = await SharedPreferences.getInstance();
await prefs.setInt(key, timestamp);
```

## Why it matters

Navigation is sequenced **behind** a persistence side effect, and the side effect is unguarded. If `SharedPreferences.getInstance()` or `setInt` throws — platform-channel failure, storage full, corrupted preferences, a plugin not yet registered — the `await` rethrows, `onTap()` is never reached, and the tap does nothing at all. No navigation, no error message, no retry: the user taps a dispute and the app appears frozen. Marking a row as read is bookkeeping and should never be able to block the user from opening it.

This is also observable in tests: the tap callback silently does not fire unless `SharedPreferences.setMockInitialValues({})` is called first, which is what put me onto it.

## Reproduction

```dart
testWidgets('tap is swallowed when preferences are unavailable', (tester) async {
  // Note: no SharedPreferences.setMockInitialValues(...) here.
  var taps = 0;
  await tester.pumpWidget(/* ... DisputeListItem(dispute: ..., onTap: () => taps++) ... */);

  await tester.tap(find.byType(GestureDetector).first);
  await tester.pumpAndSettle();

  expect(taps, 1); // fails: taps == 0
});
```

## How to fix

Navigate first, persist as a non-blocking side effect, and never let bookkeeping surface as a dead tap:

```dart
onTap: () {
  onTap();
  unawaited(
    DisputeReadStatusService.markDisputeAsRead(dispute.disputeId)
        .catchError((Object e, StackTrace s) {
      logger.w('Failed to mark dispute ${dispute.disputeId} as read', error: e, stackTrace: s);
    }),
  );
},
```

If the ordering genuinely matters (so the unread badge clears before the next screen paints), keep the `await` but wrap it so a failure cannot block navigation:

```dart
onTap: () async {
  try {
    await DisputeReadStatusService.markDisputeAsRead(dispute.disputeId);
  } catch (e, s) {
    logger.w('Failed to mark dispute as read', error: e, stackTrace: s);
  }
  onTap();
},
```

Either way, add error handling inside `DisputeReadStatusService` itself so `markDisputeAsRead` / `getLastReadTime` / `hasUnreadMessages` degrade gracefully rather than propagating storage failures into the UI.

## Related

`lib/features/disputes/widgets/dispute_content.dart` also calls into `DisputeReadStatusService` from a `FutureBuilder`; worth checking that a rejected future there renders something sensible rather than an error box.

21Mill:
Closes #653

## Problem

`DisputeListItem` sequenced navigation *behind* a persistence side effect, and the side effect was unguarded:

```dart
onTap: () async {
  await DisputeReadStatusService.markDisputeAsRead(dispute.disputeId);
  onTap();
},
```

`markDisputeAsRead` goes through `SharedPreferences.getInstance()` + `setInt`. If either throws — platform-channel failure, storage full, corrupted preferences, a plugin not yet registered — the `await` rethrows and `onTap()` is never reached. No navigation, no error message, no retry: the user taps a dispute and the app appears frozen.

Marking a row as read is bookkeeping. It should never be able to stand between the user and the dispute.

## Change

- `lib/features/disputes/widgets/dispute_list_item.dart` — the tap callback runs first; the write follows as a fire-and-forget side effect whose failure is logged rather than propagated.

```dart
onTap: () {
  onTap();
  unawaited(
    DisputeReadStatusService.markDisputeAsRead(dispute.disputeId)
        .catchError((Object e, StackTrace s) {
      logger.w('Failed to mark dispute ${dispute.disputeId} as read',
          error: e, stackTrace: s);
    }),
  );
},
```

- `lib/features/disputes/screens/dispute_chat_screen.dart` — the same method was called from the post-frame callback without `await` and without a catch, turning the same failure into an unhandled asynchronous error. It now goes through the same guard.

No other production code changed, and no user-facing strings were added.

### Considered and rejected

`/dispute_details/:id` resolves to `DisputeChatScreen`, which already marks the dispute as read in `initState`, so the list item's write is redundant *for its current caller*. I did not remove it: `onTap` is an injected callback and the widget cannot assume where its caller navigates.

Out of scope: `dispute_content.dart` builds its `hasUnreadMessages` future inside `build()`, but its `snapshot.data ?? false` already degrades gracefully if that read fails.

## Tests

`test/features/disputes/widgets/dispute_widgets_test.dart` gains a `_ThrowingStore extends SharedPreferencesStorePlatform` whose every operation throws, installed together with `SharedPreferences.resetStatic()` so the cached instance cannot mask it, and restored in `addTearDown`:

- `reports taps even when preferences are unavailable` — the regression test. Against the old widget it reproduces the reported symptom exactly: `Expected: <1>` tap, `Actual: <0>`.
- `still records the dispute as read` — with the normal mock store, tapping leaves `getLastReadTime('dispute-1')` non-null, pinning that the fix did not drop the bookkeeping.

`shared_preferences_platform_interface` is already a direct dev dependency.

## Test plan

- [x] `flutter analyze` on `lib/features/disputes` and `test/features/disputes` — no issues
- [x] `flutter test test/features/disputes/` — 24 passing
- [x] Regression test confirmed failing against the unfixed widget
- [x] Full `flutter test` — 890 passing, 11 pre-existing failures (see below)
- [ ] CI green

### Pre-existing failures, unrelated to this PR

Eleven test files fail to load because `test/mocks.mocks.dart` is stale (`MockSettings.copyWith` and `MockOrderState.copyWith` lack named parameters their overridden methods now declare). They cannot be regenerated locally: `pub get` resolves `analyzer 8.x` and `build_runner` fails to compile against it — repo issue #606. None of them touch the disputes feature, and CI regenerates mocks itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dispute navigation now opens immediately, even if marking a dispute as read fails.
  * Read-status updates continue in the background without blocking the chat or dispute list.
  * Failures during read-status updates are handled safely and recorded for troubleshooting.

* **Tests**
  * Added coverage for successful and failed dispute read-status updates.

### DIFF
--- a/lib/features/disputes/screens/dispute_chat_screen.dart
+++ b/lib/features/disputes/screens/dispute_chat_screen.dart
@@ -7,7 +7,6 @@ import 'package:mostro_mobile/features/chat/providers/active_chat_screens_provid
 import 'package:mostro_mobile/features/disputes/widgets/dispute_communication_section.dart';
 import 'package:mostro_mobile/features/disputes/widgets/dispute_message_input.dart';
 import 'package:mostro_mobile/features/disputes/providers/dispute_providers.dart';
-import 'package:mostro_mobile/services/logger_service.dart';
 import 'package:mostro_mobile/data/models/dispute.dart';
 import 'package:mostro_mobile/data/models/session.dart';
 import 'package:mostro_mobile/core/app_theme.dart';
@@ -36,16 +35,7 @@ class _DisputeChatScreenState extends ConsumerState<DisputeChatScreen> {
     // Mark dispute as read when screen is opened
     WidgetsBinding.instance.addPostFrameCallback((_) {
       if (!mounted) return;
-      unawaited(
-        DisputeReadStatusService.markDisputeAsRead(widget.disputeId)
-            .catchError((Object e, StackTrace s) {
-          logger.w(
-            'Failed to mark dispute ${widget.disputeId} as read',
-            error: e,
-            stackTrace: s,
-          );
-        }),
-      );
+      unawaited(DisputeReadStatusService.markDisputeAsRead(widget.disputeId));
       // Notify that the dispute has been marked as read
       ref.read(disputeReadStatusProvider(widget.disputeId).notifier).state =
           DateTime.now().millisecondsSinceEpoch;
--- a/lib/features/disputes/screens/dispute_chat_screen.dart
+++ b/lib/features/disputes/screens/dispute_chat_screen.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:mostro_mobile/data/enums.dart' as enums;
@@ -5,6 +7,7 @@ import 'package:mostro_mobile/features/chat/providers/active_chat_screens_provid
 import 'package:mostro_mobile/features/disputes/widgets/dispute_communication_section.dart';
 import 'package:mostro_mobile/features/disputes/widgets/dispute_message_input.dart';
 import 'package:mostro_mobile/features/disputes/providers/dispute_providers.dart';
+import 'package:mostro_mobile/services/logger_service.dart';
 import 'package:mostro_mobile/data/models/dispute.dart';
 import 'package:mostro_mobile/data/models/session.dart';
 import 'package:mostro_mobile/core/app_theme.dart';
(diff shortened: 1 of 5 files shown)


21Mill:
The Linked Issues check was right, and I had missed that line of #653: the guarantee belongs inside `DisputeReadStatusService`, not only at the two call sites. Fixed in bf67637.

- `markDisputeAsRead` logs and swallows a storage failure; `getLastReadTime` returns `null`, which every caller already treats as "nothing read yet".
- `hasUnreadMessages` reaches storage only through `getLastReadTime`, so it degrades to counting every incoming message as unread. That is stated in its doc comment rather than wrapped in a catch that could never fire — say so if you would rather see the explicit catch.
- With the guarantee stated and tested in one place, the `catchError` at both call sites was unreachable, so they are back to a plain `unawaited` call. The widget regression test still covers them: I removed the service catch and confirmed `reports taps even when preferences are unavailable` fails again, since the escaping error reaches the test binding.
- New `test/services/dispute_read_status_service_test.dart`: six tests over both a working store and one that fails every operation.

On the "Related" note in the issue: `dispute_content.dart` needs no change. Its `FutureBuilder` calls `hasUnreadMessages`, which no longer rejects on storage failure, and `snapshot.data ?? false` already renders no unread dot either way.

`flutter analyze` clean; 896 passing, with the same 11 pre-existing failures from the stale `test/mocks.mocks.dart` (#606).

Catrya (approved):
tACK

PR #669, opened 2026-08-21
feat: show a shareable link on takeable orders

### LINKED ISSUE #451: Display order link with copy option in order details
Add a section in the order details view that displays the order's shareable link with a copy button. When tapped, the link is copied to the clipboard with visual feedback (e.g., "Copied!"). This allows users to easily share that specific order with others who can then access it directly from any Mostro client.

21Mill:
## Depends on #673

#673 fixes the receiving end (#670): until it lands, a recipient opening one of these links with the app closed does not reliably reach the order. Merge order: #673 → this.

Closes #451

## Problem

There is no way to share a specific order from the app. Someone who creates an order and waits for a taker can only read its id aloud, leaving the other side to look it up by hand.

What makes this odd is that half the work was already there, pointing the other way: **the app knows how to open order links but never produces one.**

- The `mostro:` scheme is registered in `android/app/src/main/AndroidManifest.xml`.
- `NostrUtils.isValidMostroUrl` / `parseMostroUrl` define the format: `mostro:<order-id>?relays=wss://r1,wss://r2&mostro=<pubkey-hex>`, where `relays` is mandatory and `mostro` is optional but must be 64-char hex.
- `DeepLinkService.processMostroLink` resolves the order from those relays, and `DeepLinkHandler` even switches Mostro instance when the link points at a different one.

So no format had to be invented and nothing had to be agreed with other clients — only the encoder was missing.

## Change

**`NostrUtils.buildMostroUrl`**, placed right next to the two functions that read the same format, so all three stay together:

```dart
static String? buildMostroUrl({
  required String orderId,
  required List<String> relays,
  String? mostroPubkey,
  int maxRelays = 3,
})
```

It keeps only relays carrying a WebSocket scheme (same check `isValidMostroUrl` applies), drops duplicates, caps the list so the link stays pasteable, and emits the instance pubkey only when it passes the exact validation `parseMostroUrl` uses. It returns `null` when no usable relay is left, since the format cannot resolve an order without one. The invariant: **anything this builds is something the project's own parser accepts and reads back unchanged.**

**`OrderShareLinkCard`** shows the link with a copy button, following the shape of the `OrderIdCard` beside it, and renders nothing when the link is `null`. It reads the active relays and instance pubkey from `settingsProvider`.

The card on a sell order's detail screen, below the order id:

Order link card showing a mostro: link below the order id card

Copying puts the whole link on the clipboard — three relays and the instance pubkey — even though the card itself elides it:

Copy confirmation with the system clipboard preview showing the full link, relays and mostro pubkey included

It appears on the order book's detail screen, and in My Trades only while the order is still `pending`: a link to a trade already under way is of no use to whoever opens it.

Adds `orderLinkLabel` and `orderLinkCopiedMessage` to the six locales. No `OrderIdCard` refactor and no new automation id, so the automation contract is untouched.

Copying is the whole of what #451 asked for, so that is all this does. Offering the system share sheet as well is tracked separately in #671.

## Tests

- `test/shared/utils/deep_link_parsing_test.dart` — new `buildMostroUrl` group: round-trip through `isValidMostroUrl` + `parseMostroUrl`; non-WebSocket relays dropped; duplicates removed; relay count capped; `null` with no usable relay and with no order id; pubkey accepted upper-case or `0x`-prefixed, and omitted when malformed rather than emitted for the parser to discard.
- `test/shared/widgets/order_cards_test.dart` — the card renders the full link, the copy button puts exactly that string on the clipboard (asserted through `SystemChannels.platform`), and nothing renders when no relay is configured.

## Test plan

- [x] `flutter gen-l10n` — no untranslated messages
- [x] `flutter analyze` on `lib/shared` and `test/shared` — no issues
- [x] `flutter test test/shared/` — 203 passing
- [x] Full `flutter test` — 899 passing, 11 pre-existing failures (stale `test/mocks.mocks.dart`, repo issue #606; same list before and after this branch)
- [x] On device (Android 16, debug build): the card renders under the order id; copying shows the confirmation and the system clipboard preview shows the whole link — three relays plus `&mostro=…`
- [x] End-to-end: firing that generated link as a `VIEW` intent opens that exact order, so the encoder is validated against the real parser and not only in tests
- [ ] CI green

## Unrelated pre-existing bug found while testing

Opening a `mostro:` link while the app is **not** in the foreground crashes it:

```
'package:go_router/src/match.dart': Failed assertion: line 245 pos 12:
'uriPathToCompare.startsWith(newMatchedLocationToCompare)': is not true.
```

Reproduced by leaving an order detail screen with Back until the app drops out of the foreground, then opening the link. I built and installed `main` (`c3c2d7a7`) without this branch and it fails identically, so it is not introduced here — being an `assert`, release builds would not show the red screen but the navigation is just as broken.

It is worth flagging alongside this PR: until now almost nobody had a `mostro:` link to open, and this feature hands one to every user. I will file it separately.


21Mill:
Filed the pre-existing deep-link crash mentioned above as #670, with the reproduction and the check against `main`.

21Mill:
Agreed — handing out links that the recipient cannot open is the wrong order to ship this in.

The PR body now says so explicitly: this depends on #673, which fixes the receiving side (#670). Merge order is #673 → this one. No code change here; the dependency was already argued in the body, it just was not stated as a blocking one.

grunch:
Hi @21Mill this kind of UI new features needs to be discussed before implementing them, in fact we are not working on new UI features in this app, new features are going to the new app https://github.com/MostroP2P/app

please redirect this efforts to create new issues and then PRs on https://github.com/MostroP2P/app and your contributions can help us to have the new app faster

Issue #670, opened 2026-08-21
Opening a mostro: deep link while the app is not in the foreground crashes on a go_router assertion


21Mill:
## Context

Found while testing #669, which adds a shareable `mostro:` link to the order detail. The crash is **not** caused by that branch — see the verification below.

Opening a valid `mostro:` deep link while the app is not in the foreground crashes it:

```
'package:go_router/src/match.dart': Failed assertion: line 245 pos 12:
'uriPathToCompare.startsWith(newMatchedLocationToCompare)': is not true.
```

## Reproduction

On a debug build (Android 16, Nothing Phone 3a):

1. Open the app and let the order book load.
2. Tap any order to open its detail screen.
3. Press Back until the app leaves the foreground.
4. Open a `mostro:` link for a real order:

```
adb shell 'am start -a android.intent.action.VIEW -d "mostro:<order-id>?relays=wss://mostro-p2p.tech,wss://nos.lol,wss://relay.mostro.network&mostro=82fa8cb978b43c79b2156585bac2c011176a21d2aead6d9f7c575c005be88390"'
```

The app comes up on the red assertion screen instead of the order.

The link itself is fine and reaches the app intact — `logcat` shows the full URI arriving and being processed:

```
D com.llfbandit.app_links: Handled intent: action: android.intent.action.VIEW / data: mostro:<order-id>?relays=...&mostro=...
I flutter : 💡 Processing mostro link: mostro:<order-id>?relays=...&mostro=...
I flutter : #0   DeepLinkService.navigateToOrder (package:mostro_mobile/services/deep_link_service.dart:250:12)
```

The **same link opened while the app is already in the foreground works correctly** and navigates to the order, repeatedly. Only the not-in-foreground path fails, which points at the navigation done from `DeepLinkHandler.handleInitialDeepLink` / `DeepLinkService.navigateToOrder` running against a router whose location is not what the assertion expects.

## Not caused by #669

Verified by building and installing `main` at `c3c2d7a7` with that branch checked out nowhere near it, then repeating steps 1-4: identical assertion, identical screen.

## Why it matters

`assert` is stripped from release builds, so users would not see the red screen — but the navigation is just as broken, and the failure mode there is silent instead of loud.

This has been low-impact so far because almost nobody had a `mostro:` link to open. #669 puts one behind a copy button on every takeable order, so links are about to become common: someone pastes one into Telegram, the recipient taps it while the app is backgrounded, and this is what they get.

Issue #654, opened 2026-08-16
OrderFilter overflows horizontally by 41 px inside its fixed 320 px panel


grunch:
## Context

Found while writing widget tests for the order filter (PR #651).

Rendering `OrderFilter` (`lib/shared/widgets/order_filter.dart`) raises:

```
A RenderFlex overflowed by 41 pixels on the right.
```

## Why it matters

The panel hardcodes its own width:

```dart
// lib/shared/widgets/order_filter.dart
return Container(
  width: 320,
  height: MediaQuery.of(context).size.height * 0.8,
  padding: const EdgeInsets.all(20),
  ...
```

With `width: 320` and `padding: 20` on each side, children get 280 logical pixels. One of the inner `Row`s needs 321. Because the width is fixed rather than derived from the screen, this is **device-independent**: it overflows the same 41 px on a small phone and on a tablet. In debug builds users see the yellow/black overflow stripes; in release the content is silently clipped, so part of a filter control is unreachable.

## Reproduction

```dart
testWidgets('overflows', (tester) async {
  await tester.pumpWidget(
    ProviderScope(
      overrides: [
        currencyCodesProvider.overrideWith((ref) async => someCurrencies),
        paymentMethodsDataProvider.overrideWith((ref) async => somePaymentMethods),
      ],
      child: const MaterialApp(home: Scaffold(body: Center(child: OrderFilter()))),
    ),
  );
  await tester.pump();

  expect(tester.takeException(), isNull); // fails: RenderFlex overflowed by 41 pixels
});
```

Current behaviour is pinned in `test/shared/widgets/order_filter_test.dart`:

```dart
// The panel is pinned to a 320 px width while one of its rows needs more
// room, so Flutter reports a horizontal overflow.
testWidgets('currently overflows horizontally', (tester) async { ... });
```

and the other tests in that file tolerate the overflow through an
`expectNoUnexpectedError` helper.

## How to fix

Find the offending `Row` (start from the rating / premium range rows and the days row — the ones combining a label, a slider or range control and a value) and pick one of:

1. Wrap the flexible child in `Expanded` / `Flexible` so it yields instead of demanding its intrinsic width.
2. Replace the `Row` with a `Wrap` where the label and control can fall onto two lines.
3. Drop the hardcoded `width: 320` in favour of a width derived from the available constraints, e.g.
   `width: math.min(320, MediaQuery.of(context).size.width - 32)`, so narrow devices are handled too.

## Follow-up

[...]

PR #672, opened 2026-08-21
feat: share an order through the system share sheet

### LINKED ISSUE #671: Share an order through the system share sheet, not only by copying its link
## Context

#669 (closing #451) adds a shareable `mostro:` link to the order detail with a copy button. Copying was the whole of what #451 asked for, and the scope was kept to exactly that on purpose — this issue is the deliberate follow-up, not an oversight.

## Proposal

Alongside the copy button, offer the system share sheet, so an order can go straight to Telegram, Signal, WhatsApp, a note, or anything else the user has installed — without the copy-switch-app-paste detour.

## Why it is cheap

Everything it needs already exists:

- `share_plus` is already a direct dependency (`pubspec.yaml:80`), so nothing new is pulled in.
- `NostrUtils.buildMostroUrl` already produces the link, and `OrderShareLinkCard` already has it in hand.

The change is a second `IconButton` in that card, next to the copy one, calling the share sheet with the same string. Worth deciding as part of it:

- Whether to share the bare link or a short sentence around it ("Take this order on Mostro: <link>"). A sentence reads better in a chat, but it is user-facing copy and would need the six locales.
- Whether the share sheet replaces the copy button or sits beside it. Keeping both seems right: copy is faster when the destination is not an app (a password manager, a terminal, a QR generator).

## Note

Sharing is only offered where the link itself is offered: on the order book's detail screen, and in My Trades only while the order is still `pending`. Whatever is decided here should keep that rule — a link to a trade already under way is of no use to whoever receives it.

21Mill:
## Depends on #669 and #673

Cut from `feat/order-share-link`, which is where `OrderShareLinkCard` lives. Until #669 merges this diff carries its commits too — review against that branch, or merge #669 first.

#673 fixes the receiving end (#670): until it lands, a recipient opening one of these links with the app closed does not reliably reach the order. Merge order: #673 → #669 → this.

Closes #671

## Problem

#669 put the order's `mostro:` link on the detail screen with a copy button, which was the whole of what #451 asked for. Copying still means copy → leave the app → open the chat → paste.

There is also something the clipboard cannot do: send the link with an explanation. A bare `mostro:8927bb1d-…?relays=…` lands in a chat as an unexplained string, and most apps only linkify `http(s)`, so the recipient gets neither a tappable link nor a hint of what they are looking at.

## Change

A share button beside the copy one in `OrderShareLinkCard`, handing the link to the system share sheet wrapped in a sentence: **"Take this order on Mostro: `<link>`"**.

Both actions stay. Copying is still faster when the destination is not an app — a password manager, a terminal, a QR generator.

The share call follows the pattern `pay_lightning_invoice_widget.dart` already uses: everything it needs is read from the context *before* the await, since this card is a `ConsumerWidget` and has no `mounted` to check afterwards, and a failure is reported through the snackbar instead of escaping. It passes `sharePositionOrigin`, which iPads need to anchor the sheet; the three existing `Share.share` calls in the repo omit it, and #80 will eventually want it.

Adds `orderShareMessage` and `failedToShareOrder` to the six locales. `share_plus` was already a direct dependency, so nothing new is pulled in.

The card now carries both actions, with the link still elided to two lines:

Order link card with the copy and share buttons side by side

Tapping share opens the system sheet with the link whole — sentence, three relays and the instance pubkey — which is what that elision hides:

System share sheet preview showing the sentence followed by the complete mostro link

## Tests

`test/shared/widgets/order_cards_test.dart`, extending the group #669 added:

- Tapping share invokes `dev.fluttercommunity.plus/share` with text containing the full link, and asserts the text does **not** start with `mostro:` — the sentence is the point, so its absence should fail the test.
- The share button disappears along with the rest of the card when no relay can resolve the order.
- A layout test: the card now holds two 48 dp buttons next to a link long enough to wrap, which is exactly the shape that produced the overflow in #654. **Confirmed the test fails** (`RenderFlex overflowed by 152 pixels`) against a card whose link is not flexible, so it is a real guard and not one that passes by construction.

## Test plan

- [x] `flutter gen-l10n` — no untranslated messages
- [x] `flutter analyze` on `lib/shared` and `test/shared` — no issues
- [x] `flutter test test/shared/widgets/order_cards_test.dart` — 18 passing
- [x] Full `flutter test` — 901 passing, the same 11 pre-existing failures as on `main` (stale `test/mocks.mocks.dart`, repo issue #606)
- [x] On device (Android 16, debug build): both buttons render without disturbing the link, the share sheet opens with the complete text, and copying still behaves as before
- [ ] CI green

## Known limitation

A shared `mostro:` link will not be tappable in most chat apps, so the recipient has to copy it and open it. Fixing that properly needs verified `https` links (App Links / Universal Links) redirecting to the scheme, which is a separate piece of work — worth its own issue if the maintainers want it.

Note also #670: opening a `mostro:` link while the app is not in the foreground currently crashes on a go_router assertion. Pre-existing and unrelated to this branch, but this feature makes such links common.


21Mill:
Agreed. The body now states both dependencies explicitly: this waits on #669 (where `OrderShareLinkCard` lives) and on #673, which fixes the receiving side (#670). Merge order: #673 → #669 → this one.

No code change here — the share sheet only makes the same links easier to send, so the fix belongs upstream in #673 rather than in this stack.

grunch:
same comment https://github.com/MostroP2P/mobile/pull/669#issuecomment-5395482381 

I'm closing this and all other UI PRs, we only are fixing bugs on this apps, not new features, new features in https://github.com/MostroP2P/app please

Issue #11, opened 2024-05-08
Create a private chat for buyer and seller


Catrya:
After users starts to talk to each other the client should make them chat sending nostr events with kind 14 and the content encrypted with nip44 algorithm, this messages leaks metadata exactly the same like nip04, but with nip44 we have the option of allow users to export their conversation key and voluntarily gave it to an admin if they need, having the opportunity of read the messages from both parties the admin can easily solve the dispute.

But if a user uses the same key for all trades the conversation key would be the same if the trade each other again, so the admin could see older messages, that's why we should force users to use a ephemeral key for each trade.

This goal of this issue is to create a chat between parties after the order is active to exchange directly how the buyer will send the fiat money to the seller.

Must adhere to [this design](https://github.com/MostroP2P/mobile/discussions/23).

bilthon:
I have a couple of questions. First, since the shared key is a point, how exactly do we apply the Lowent method to convert it into a private key? I see that a hash is used, so the question boils down to whether we hash the compressed or uncompressed version of the shared secret 𝑘.

The second question is about the mechanics of the Lowent chat. Here, both parties retain their own Nostr private keys but send messages to the shared "room" created by their shared secret. To an external observer, it would appear as if both Alice and Bob are sending messages to this third unknown entity. Is that correct?

Catrya:
> The second question is about the mechanics of the Lowent chat. Here, both parties retain their own Nostr private keys but send messages to the shared "room" created by their shared secret. To an external observer, it would appear as if both Alice and Bob are sending messages to this third unknown entity. Is that correct?

Alice and Bob will share a new key and use that key to send DMs to that same key. In the eyes of an observer, a random user will be sending a DM to theyself, and will not be able to link with either Alice or Bob.

bilthon:
But in that case the client won't be able to distinguish between messages from each party and provide a proper chat UX like mostro web does today.

Catrya:
Well each user will know what they wrote, since it will only be a chat between 2 people I think it should not generate so much confusion among users, they will just not see their handle placed in the message

gzuuus:
> I have a couple of questions. First, since the shared key is a point, how exactly do we apply the Lowent method to convert it into a private key? I see that a hash is used, so the question boils down to whether we hash the compressed or uncompressed version of the shared secret 𝑘.

[...]

Issue #10, opened 2024-05-08
Connecting the buyer and the seller


Catrya:
References:  https://mostro.network/messages/seller_pay_hold_invoice.html
https://mostro.network/messages/take_buy.html#buyer-sends-ln-invoice

After the seller paid hold invoice, and the buyer already sent the invoice to receive the sats, Mostro will send a new message to both parties containing each other's public keys to initiate a direct conversation.
Also Mostro updates the event 38383 about that order to change the status to `active`

grunch:
closing by duplication with #11

PR #673, opened 2026-08-21
fix: do not let a cold-start deep link become the router's location

### LINKED ISSUE #670: Opening a mostro: deep link while the app is not in the foreground crashes on a go_router assertion

21Mill:
Closes #670

## Problem

Opening a `mostro:` link while the app was not running crashed it before anything rendered:

```
'package:go_router/src/match.dart': Failed assertion: line 245 pos 12:
'uriPathToCompare.startsWith(newMatchedLocationToCompare)': is not true.
```

With no activity alive, Android does not deliver the link through `pushRouteInformation`: it hands it over as the engine's `defaultRouteName`. And go_router prefers that over `initialLocation` whenever it is not `/` (`go_router-16.0.0/lib/src/router.dart:546-571`):

```dart
if (initialLocation == null)      return platformDefault;
else if (platformDefault == '/')  return initialLocation;
else                              return platformDefault;   // <- here
```

So the router started up trying to match `mostro:<id>?relays=…`. That is an *opaque* URI: its `path` is the bare id with no leading slash, so `'8927…'.startsWith('/')` is false and the assertion fires.

This also explains why the two existing guards did not help. `DeepLinkInterceptor` covers the `pushRouteInformation` path, which is not the one used here. And the redirect in `app_routes.dart` that sends custom schemes home never runs, because matching asserts before redirects are consulted.

## Change

`createRouter` sets `overridePlatformDefaultLocation` when the platform default carries one of our schemes, so the app starts at `/` and the initial link is left to `_processInitialDeepLink` in `MostroApp`, which reads it through `app_links`.

The override is **conditional** rather than always on: on web the platform default is the location the user actually asked for, and discarding it would break opening the app at a URL. The discarded value is now logged, so a link that goes missing on a cold start leaves a trace.

The "is this one of our schemes" test already existed twice — in the interceptor and in the redirect — and this would have added a third copy, so it now lives on its own in `lib/core/deep_link_schemes.dart`. It is not in the interceptor: `app_routes.dart` importing it from there would point the dependency the wrong way, since routing is the lower layer and the interceptor is one of its consumers (@grunch's point in review).

## Scope

**This PR fixes the crash. It does not change delivery.** A cold start link still reaches the app the way it does on `main` — a post frame callback plus a `100 ms` delay, and a silent drop if the router is not up yet. On the device that gap is real: the link needs about **0.9 s** before the app can open it. So after this PR the app no longer crashes, but on a slow start the link may still fail to open the order, exactly as before.

That delivery layer is #692, which sits on top of this branch. This one stands on its own and is worth merging first: it removes the crash, and #669 and #672 are waiting on it.

## Tests

`TestPlatformDispatcher.defaultRouteNameTestValue` lets the cold start be reproduced without a device.

- `test/core/app_routes_test.dart` — a custom scheme handed over by the platform is ignored and the router starts at `/`; an ordinary launch starts at `/`; a real location like `/settings` still wins, which is the web case. **Confirmed the first test fails against the unfixed router**, reporting the initial location as `mostro:8927bb1d-…` itself — the defect exactly, not a proxy for it. The helper now holds a single router for the test and disposes it, instead of building one inside a `Consumer.builder` that may run more than once.
- `test/core/deep_link_schemes_test.dart` — `isCustomSchemeLocation` over `mostro:`, `lightning:`, app locations, http(s) and unparseable input.

## Test plan

- [x] `flutter analyze` on `lib/core` and `test/core` — no new issues (two pre-existing `containsSemantics` deprecation infos in `automation_contract_test.dart`)
- [x] `flutter test test/core/` — 52 passing
- [x] Full `flutter test` — 897 passing, the same 11 pre-existing failures as on `main` (stale `test/mocks.mocks.dart`, #606)
- [x] On device (OnePlus 8T, Android 14, debug build), all three paths:
  - **cold start with a link** (the one that crashed): opens that order, no assertion
  - **app in the foreground**: still opens the order, as before
  - **plain launch, no link**: opens the order book
- [ ] CI green

## Note

Found while testing #669, which puts a `mostro:` link behind a copy button on every takeable order — so links are about to become common. This PR is independent of that one and of #672, and applies to `main` on its own.


21Mill:
Fixed in the pushed commit. `startsWith('http')` came from the predicate this consolidates — it was already written that way in both the interceptor and the redirect — so the looseness rode along into the shared helper.

Now compared exactly against `http` and `https`, with a regression test for `httpfoo://example.com`. I did not add a `toLowerCase()`: `Uri` already normalises the scheme, so there is a test asserting `HTTPS://…` is still treated as web rather than defensive code that cannot change the outcome.

21Mill:
Good catch, and you are right: the review found a real hole that my own device testing had hidden.

Once `createRouter()` discards the platform default, the cold-start link depends entirely on `_processInitialDeepLink()`. That path waited one frame plus a fixed 100 ms and then gave up if `_router` was still `null` — and `_router` is only created in the `data` branch of `appInitializerProvider`, so while Nostr/keys/sessions are still starting up the loading `MaterialApp` is on screen and there is no router to hand the link to. On my phone initialization won that race, which is all my manual test proved.

`4d1a3fd1` replaces the timing guess with a stored link:

- `lib/core/initial_deep_link_queue.dart` — a small `InitialDeepLinkQueue` that holds the URI and delivers it exactly once, and only when a router is passed in.
- `lib/core/app.dart` — `_processInitialDeepLink()` stores the link and `_drainInitialLink()` hands it over in a post-frame callback; the same drain runs right after `_router ??= createRouter(ref)`. The 100 ms delay is gone. A slow start now delays the order screen instead of dropping the link.

On the test: I pulled the coordination into its own class precisely so the delayed-router case could be tested, and `test/core/initial_deep_link_queue_test.dart` covers it — link stored, drained with no router (not delivered, still pending), drained again once the router exists (delivered), plus delivered-only-once. I did not write a widget test around `MostroApp` because pumping it means standing up the whole provider graph (`appInitializerProvider`, settings, auth, community, lifecycle) and mocking the `app_links` channel; the value would be in the wiring, and the cost in flakiness. Happy to add it if you would rather have it.

`flutter analyze` is clean on the touched files and `flutter test test/core/` is green (55 tests). The full suite still shows the same 11 pre-existing load failures from the missing generated mocks (#606). I could not re-run the device check this time — the phone is not connected right now — so the on-device evidence is still the one from the earlier comment, which covers the crash but not the slow-init case.

21Mill:
Applied in `f1782ac8`: both error paths around the initial deep link now go through the `logger` singleton (`logger.e` with `error`/`stackTrace`) instead of `debugPrint`.

Note that one of the two, `Error processing initial deep link`, is pre-existing code — I only moved the other one out of the deleted `_handleInitialMostroLink`. I converted both since they sit in the same flow, but I left the rest of `app.dart` alone: it still uses `debugPrint` in seven other places, and rewriting those here would be unrelated churn in a bugfix PR. Worth a separate pass if you want the file consistent.

`flutter analyze` clean, `flutter test test/core/` green (55 tests).

grunch (changes requested):
## Review: Request changes

The diagnosis is correct and `overridePlatformDefaultLocation` is the right fix. My concern is commit 3 (`InitialDeepLinkQueue`): it only partially resolves the blocker from the previous review, and it reintroduces the same silent-drop in the opposite window.

**Verification performed locally** (this repo, PR head `f1782ac`):
- `flutter analyze lib/core test/core` — 2 pre-existing `containsSemantics` infos only, no new issues.
- `flutter test test/core/app_routes_test.dart test/core/deep_link_interceptor_test.dart test/core/initial_deep_link_queue_test.dart` — 12/12 passing.
- Read `go_router-16.0.0/lib/src/router.dart:546-571` and `packages/flutter/lib/src/scheduler/binding.dart:788-802` from the resolved SDK/lockfile.

---

### 🔴 HIGH-1 — `addPostFrameCallback` does not schedule a frame, so the link can still be dropped

`lib/core/app.dart:101-113`

```dart
void _drainInitialLink() {
  if (!_initialLink.isPending) return;
  WidgetsBinding.instance.addPostFrameCallback((_) async { ... });
}
```

From `scheduler/binding.dart:793-797`:

> This method does *not* request a new frame. […] Otherwise, the registered callback is executed after the next frame (**whenever that may be, if ever**).

Concrete failure path: `appInitializerProvider` resolves quickly (session already restored, relays cached), the `data` branch builds, the router is created, `_drainInitialLink()` runs with `isPending == false` and returns. The UI settles and stops requesting frames. *Then* `appLinks.getInitialLink()` (line 84) resolves → `store()` + `_drainInitialLink()` → a post-frame callback is registered with no frame scheduled → the callback never runs and the order never opens.

This is not hypothetical: line 92 executes inside a `Future` continuation with no frame in progress by construction. It is the same defect this PR exists to remove, just in the opposite timing window.

Minimal fix — request a frame so the callback is guaranteed to run:

```dart
WidgetsBinding.instance.addPostFrameCallback((_) async { ... });
WidgetsBinding.instance.ensureVisualUpdate();
```

(or deliver synchronously when `_router` is already non-null).

---

### 🟠 MEDIUM-1 — The requested regression test is still missing

`test/core/initial_deep_link_queue_test.dart` exercises `InitialDeepLinkQueue` in isolation, and that class never had the bug. The defect lived in the `_MostroAppState` wiring: that `_drainInitialLink()` is invoked after `_router ??= createRouter(ref)` (`app.dart:152`) and survives the `loading → data` transition.

No test mounts `MostroApp` with `appInitializerProvider` in `loading`, resolves it to `data`, and asserts the link is delivered. Deleting line 152 entirely leaves all three new tests green — the test does not protect the fix.

The earlier request was specifically for "a regression test that covers the delayed-router case"; that is still open.

---

### 🟠 MEDIUM-2 — `drain()` clears the pending link before delivering it

`lib/core/initial_deep_link_queue.dart:16-20`

```dart
final uri = _pending;
if (uri == null || router == null) return;
_pending = null;          // cleared BEFORE delivery
await deliver(uri, router);
```

If `deliver` throws — `app.dart:108` catches and only logs — the link is already discarded and there is no retry. In a change whose stated goal is "a slow start delays the order screen rather than losing it", the error path does the opposite. Clear `_pending` only after a successful `await`, or restore it in the `catch`.

---

### 🟠 MEDIUM-3 — The same bug class is left unfixed 40 lines above

`lib/core/app.dart:64`

```dart
_customUrlSubscription = _deepLinkInterceptor!.customUrlStream.listen(
  (url) async {
    if (_router != null) { ... }   // dropped silently when null
  },
```

A link delivered through `didPushRouteInformation` while initialization is still in flight (process alive, `MostroApp` freshly mounted) is lost in exactly the same way. This PR introduces a reusable abstraction for precisely this and does not apply it here. If `InitialDeepLinkQueue` is the right answer, this branch should use it.

---

### 🟡 LOW-1 — Logger migration is incomplete

Commit 4 replaces 2 `debugPrint` calls, but 7 remain in `app.dart` (lines 61, 70, 75, 87, 160, 161, 176) — including line 87, inside the very method being changed. The file now mixes two logging conventions. Note `debugPrint` is not stripped in release, and line 87 dumps the full link (order id + relays).

### 🟡 LOW-2 — Version reference in the description does not match the lockfile

The PR body cites `go_router-17.1.0/lib/src/router.dart:630-649`. `pubspec.lock` pins **16.0.0**, where the block is at `router.dart:546-571`. The logic is identical and the analysis holds, but the citation is not reproducible against this repo.

### 🟡 LOW-3 — `InitialDeepLinkQueue` is not a queue

It holds a single `Uri` and `store()` overwrites silently without signalling the discard. It also imports `go_router` solely for a parameter type, coupling a trivial holder to the router. A `Uri?` field on the `State` plus a `_deliverPendingLink()` method would cover the same ground without a new file or class — and would be equally testable if the test were at the widget level (see MEDIUM-1).

### 🟡 LOW-4 — Asymmetry between what is discarded and what is handled

`isCustomSchemeUri` claims **any** non-http(s) scheme, so `createRouter` discards the platform default for e.g. `lightning:`. But `_processInitialDeepLink` (line 86) only stores `scheme == 'mostro'`. Any custom scheme other than `mostro:` is discarded from the router *and* left unhandled. This is theoretical today — I verified `AndroidManifest.xml` declares only `mostro` as an inbound `intent-filter` (`lightning` is under `<queries>`, outbound) and `Info.plist` lists only `mostro` — but the `'claims other non-web schemes'` test asserts a capability the app does not actually have.

---

### ✅ What is right

- The root-cause analysis is correct and verifiable: in go_router 16.0.0's `_effectiveInitialLocation`, a `platformDefault != '/'` wins over `initialLocation`, and `Uri.parse('mostro:8927…')` is opaque (`hasEmptyPath == false`), so it does not get normalised to `/`.
- Making `overridePlatformDefaultLocation` **conditional** is the right call: `web/` exists in this repo, and forcing it unconditionally would break opening the app at a URL. go_router's assert at `router.dart:191` requires `initialLocation != null`, which is satisfied.
- The exact `http`/`https` match in commit 2 is correct, and pinning the `HTTPS://` case against `Uri`'s scheme normalisation rather than adding a defensive `toLowerCase` is the better choice.
- Collapsing the predicate into one place instead of adding a third copy is a genuine improvement.
- `defaultRouteNameTestValue` is the right way to reproduce the cold start without a device, and the `isCustomSchemeLocation` cases are thorough.

---

### To unblock

1. **HIGH-1** — `ensureVisualUpdate()` (or direct delivery when the router already exists).
2. **MEDIUM-1** — a `testWidgets` that mounts `MostroApp` with `appInitializerProvider` in `loading`, resolves it to `data`, and asserts the link is delivered exactly once.
3. **MEDIUM-2** — do not discard the link on the error path.

MEDIUM-3 and the LOW items are your call; MEDIUM-3 is worth doing because it is literally the same bug in a file this PR already touches.

21Mill:
Thanks for the depth here — HIGH-1 and MEDIUM-1 were both real, and I verified each of your claims against the resolved SDK before touching anything. `a0e4207d`:

### HIGH-1 — fixed, and the fix is not the one you suggested

You are right that `addPostFrameCallback` requests nothing (`scheduler/binding.dart:792-797`), and the failure path you describe is exact. I did not use `ensureVisualUpdate()` alone, because the delivery does not actually need *a frame* — it needs the **navigator**, which `DeepLinkHandler._handleMostroDeepLink` reads via `router.routerDelegate.navigatorKey.currentContext` (`deep_link_handler.dart:88-104`) and without which it logs and returns. So the condition is now that, not timing:

```dart
if (router.routerDelegate.navigatorKey.currentContext == null) {
  WidgetsBinding.instance.addPostFrameCallback((_) { if (mounted) _deliverPendingDeepLink(); });
  if (SchedulerBinding.instance.schedulerPhase == SchedulerPhase.idle) {
    // A post frame callback does not request a frame on its own.
    WidgetsBinding.instance.ensureVisualUpdate();
  }
  return;
}
_pendingDeepLink = null;
unawaited(_handleDeepLink(uri, router));
```

Your late-arrival case now delivers immediately, with no frame involved at all. `ensureVisualUpdate()` stays for the case where the router exists but the navigator is not mounted while the app is idle; it is guarded on `schedulerPhase == idle` so the retry path cannot request a frame from inside `postFrameCallbacks` and spin.

### MEDIUM-1 — done, and checked by mutation

`test/core/app_deep_link_test.dart` mounts the real `MostroApp` with `appInitializerProvider` under a `Completer`, the `app_links` channel mocked, and a recording `DeepLinkHandler`. Four tests, and I ran them against the code they are supposed to protect:

- against `f1782ac8`: **3 of 4 fail** (the late-arrival, the failed-delivery and the intercepted-stream cases).
- deleting the `_deliverPendingDeepLink()` call after `_router ??= createRouter(ref)` — the exact mutation you used: the first test fails too. Your point stands as written; it now does not.

One caveat worth stating: the completers must be created inside the test body, not in `setUp`, or they never complete for the widget under the fake async zone. Cost me an hour.

### MEDIUM-2 — fixed

Delivery happens through a helper that restores the link on failure (`_pendingDeepLink ??= uri`) and logs. The test fails the first delivery and asserts the next rebuild still finds it.

### MEDIUM-3 — fixed

`customUrlStream` no longer drops when `_router == null`; it goes through the same `_queueDeepLink`. Duplicates are already absorbed by `DeepLinkHandler`'s 2 s window on `_lastHandledDeepLinkUrl`, so a link arriving through both paths is handled once.

### LOW items

- **LOW-1**: all 7 remaining `debugPrint` calls are gone; `grep -n debugPrint lib/core/app.dart` is empty. On the link being dumped: `DeepLinkInterceptor` (`logger.i`, lines 23 and 27) and `DeepLinkHandler:50` already log the full URI, so redacting only here would change nothing real — but `logger` only reaches the console when `Config.isDebug` (`logger_service.dart:280-283`), so the release exposure you flagged is closed. A redaction policy across the three sites is worth its own change.
- **LOW-2**: PR body now cites `go_router-16.0.0/lib/src/router.dart:546-571`. Confirmed against the lockfile.
- **LOW-3**: `InitialDeepLinkQueue` is deleted, along with its test. A `Uri?` on the state plus `_deliverPendingDeepLink()`, as you proposed.
- **LOW-4**: `_processInitialDeepLink` now queues any custom scheme via `DeepLinkInterceptor.isCustomSchemeUri`, matching exactly what `createRouter` discards, and the same predicate the intercepted path already used. A `lightning:` link would now reach `DeepLinkHandler`, which answers with the localized `unsupportedLinkFormat` instead of vanishing.

### Verification

- `flutter analyze lib/core test/core` — only the 2 pre-existing `containsSemantics` infos.
- `flutter test test/core/` — 56 passing.
- `flutter test` — 901 passing, the same 11 pre-existing load failures from the missing generated mocks (#606).
- No device run: the phone is not connected right now, so the on-device evidence is still the earlier one, which covers the crash but not these timing windows.

test/core/initial_deep_link_queue_test.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:go_router/go_router.dart';
+import 'package:mostro_mobile/core/initial_deep_link_queue.dart';
+
+void main() {
+  final router = GoRouter(routes: []);
```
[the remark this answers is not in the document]
21Mill: That fixture is gone: `InitialDeepLinkQueue` and its test were removed in `a0e4207d`, replaced by a widget test that mounts the real `MostroApp`.

For the record, though, the assert you cite does not exist in go_router 16.0.0. `GoRouter`'s constructor asserts only on `initialExtra`, `overridePlatformDefaultLocation` and the `onException`/`errorBuilder` exclusivity (`router.dart:187-198`), and `RouteConfiguration` asserts paths, duplicate parameters and shell branch defaults (`configuration.dart:195-200`) — none of which require a non-empty list. `assert(routes.isNotEmpty)` is at `route.dart:668`, on `ShellRoute`/`StatefulShellBranch`, not on the router. The three tests using that fixture were passing locally with assertions on.

21Mill:
Device verification, now that I could get the phone connected again (OnePlus 8T, Android 14, debug build of `a0e4207d`, live order `a2561e55-d6d4-4d7c-9ce5-5a25378efac1`).

**Cold start** — `am force-stop` first, so the process is really gone:

```
20:11:40.464  Initial deep link detected: mostro:a2561e55-…?relays=wss://relay.mostro.network
20:11:41.402  Handling deep link: mostro:a2561e55-…
20:11:43.107  Navigating to: /take_sell/a2561e55-… (Order: a2561e55-…, Type: sell)
```

The ~0.9 s between detection and handling is the link sitting in `_pendingDeepLink` while initialization finishes — the window that used to drop it. It opened that order, not the order book, and there is no assertion in the log.

**App in foreground** — the interceptor path, which is the window HIGH-1 was about, since the navigator is already mounted and nothing is asking for frames:

```
20:12:12.370  Intercepted custom URL: mostro:a2561e55-…
20:12:12.370  Handling deep link: mostro:a2561e55-…
20:12:12.493  Navigating to: /take_sell/a2561e55-…
```

Delivered in the same millisecond it arrived. Both paths fired here (initial link and interceptor) and `DeepLinkHandler`'s 2 s window collapsed them into one navigation, as expected.

**Plain launch, no link** — order book, no deep-link log lines at all.

Not covered on device: a non-`mostro:` custom scheme (LOW-4). `AndroidManifest.xml` only registers `mostro` as an inbound filter, so there is no way to deliver one; that path is covered by the unit tests on `isCustomSchemeUri` only.

lib/core/app.dart
```diff
@@ -80,33 +78,52 @@ class _MostroAppState extends ConsumerState<MostroApp> {
...
+  }
+
+  Future<void> _handleDeepLink(Uri uri, GoRouter router) async {
+    try {
+      await ref.read(deepLinkHandlerProvider).handleInitialDeepLink(uri, router);
+    } catch (e, stack) {
+      // Keep the link so a later attempt can still open it.
+      _pendingDeepLink ??= uri;
+      logger.e('Error handling deep link', error: e, stackTrace: stack);
     }
```
grunch: 🔴 **Potential issue** | _Critical_

**The pending link does not survive a failed delivery — it is dropped silently.**

`_pendingDeepLink` is cleared *before* handing over, and is only restored if `handleInitialDeepLink` **throws**. It doesn't. `DeepLinkHandler._handleDeepLink` wraps everything in `try { … } catch (e) { logger.e(…); _showErrorSnackBar(…) }` (`lib/core/deep_link_handler.dart:47-67`) and returns normally on every real failure:

- order not found / relay down / `processMostroLink` timeout
- unsupported scheme (`logger.w` + snackbar, no throw)
- `processingContext == null` → bare `return` (`deep_link_handler.dart:99-103`)
- **user declines the Mostro-switch dialog** → `return` (`deep_link_handler.dart:126-129`)

So in production the `catch` on line 124 is effectively unreachable and the link is discarded on the first failed attempt. The behaviour the commit message describes exists only against the test double.

Related, same root cause: `_handleMostroDeepLink`'s concurrency guard (`deep_link_handler.dart:80-83`) also returns early with just a `logger.i`. Tap a second link while the first one's loading dialog is up and it is gone — the pending slot was already cleared.

**Suggested fix:** make retention depend on a *result*, not an exception — have `handleInitialDeepLink` return `Future<bool>` (or a `DeepLinkResult`) and keep the link when it reports failure. If that's out of scope for this PR, please drop `_pendingDeepLink ??= uri` and its test instead, since together they document a protection that isn't there.

lib/core/app.dart
```diff
@@ -80,33 +78,52 @@ class _MostroAppState extends ConsumerState<MostroApp> {
...
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (mounted) _deliverPendingDeepLink();
+      });
+      if (SchedulerBinding.instance.schedulerPhase == SchedulerPhase.idle) {
+        // A post frame callback does not request a frame on its own.
+        WidgetsBinding.instance.ensureVisualUpdate();
       }
-    } catch (e) {
-      debugPrint('Error handling initial mostro link: $e');
+      return;
```
grunch: 🟠 **Potential issue** | _High_

**The `== idle` guard is narrower than Flutter's own contract — this is the same failure mode commit 5 exists to remove.**

From `flutter/lib/src/scheduler/binding.dart:906-917`, `ensureVisualUpdate()` calls `scheduleFrame()` for **`idle` *and* `postFrameCallbacks`**:

```dart
void ensureVisualUpdate() {
  switch (schedulerPhase) {
    case SchedulerPhase.idle:
    case SchedulerPhase.postFrameCallbacks:
      scheduleFrame();
      return;
    case SchedulerPhase.transientCallbacks:
    case SchedulerPhase.midFrameMicrotasks:
    case SchedulerPhase.persistentCallbacks:
      return;
  }
}
```

When `_deliverPendingDeepLink()` re-enters **from its own post-frame callback** (navigator still null), the phase is `postFrameCallbacks`: a callback is registered for the next frame and **no frame is requested**. If nothing else asks for one, it never runs — precisely *"addPostFrameCallback does not request a frame"* from your own commit message.

Hard to hit today, since the `Navigator` normally mounts in the same frame as the router, so I'd call this plausible rather than confirmed. But the condition buys nothing: `ensureVisualUpdate()` already no-ops during `transientCallbacks` / `midFrameMicrotasks` / `persistentCallbacks`.

```suggestion
    if (router.routerDelegate.navigatorKey.currentContext == null) {
      WidgetsBinding.instance.addPostFrameCallback((_) {
        if (mounted) _deliverPendingDeepLink();
      });
      // A post frame callback does not request a frame on its own;
      // ensureVisualUpdate no-ops when one is already in flight.
      WidgetsBinding.instance.ensureVisualUpdate();
      return;
    }
```

This also lets you drop the `package:flutter/scheduler.dart` import added at the top of the file.

lib/core/app.dart
```diff
@@ -80,33 +78,52 @@ class _MostroAppState extends ConsumerState<MostroApp> {
...
-  /// Handle initial mostro link after router is ready
-  Future<void> _handleInitialMostroLink(Uri uri) async {
-    try {
-      // Wait for router to be ready
-      await Future.delayed(const Duration(milliseconds: 100));
+  /// Keep the link until there is a router and a navigator to open it with
+  void _queueDeepLink(Uri uri) {
+    _pendingDeepLink = uri;
+    _deliverPendingDeepLink();
+  }
```
grunch: 🟡 **Refactor suggestion** | _Medium_

**A pending link is overwritten with no trace.**

`_pendingDeepLink = uri` is unconditional. Two distinct links before the router exists — or one arriving while another delivery is in flight — and the first is lost with nothing in the log. A single slot is a defensible design, but it should say so out loud:

```suggestion
  /// Keep the link until there is a router and a navigator to open it with
  void _queueDeepLink(Uri uri) {
    if (_pendingDeepLink != null && _pendingDeepLink != uri) {
      logger.w('Replacing pending deep link $_pendingDeepLink with $uri');
    }
    _pendingDeepLink = uri;
    _deliverPendingDeepLink();
  }
```

lib/core/app.dart
```diff
@@ -80,33 +78,52 @@ class _MostroAppState extends ConsumerState<MostroApp> {
...
+
+  void _deliverPendingDeepLink() {
```
grunch: 🟡 **Potential issue** | _Medium_

**Can reach `ref.read` after `dispose`.**

`_processInitialDeepLink()` is started unawaited from `initState` (line 47). If the widget unmounts while `getInitialLink()` is still pending and `_router` had already been assigned, the flow runs straight through to `ref.read(deepLinkHandlerProvider)` on a disposed `ConsumerState`. Unlikely for the root widget in production, but a classic flake source in widget tests.

The existing `mounted` check only covers the post-frame callback; an early guard closes the entry path too:

```dart
  void _deliverPendingDeepLink() {
    if (!mounted) return;
    final uri = _pendingDeepLink;
```

lib/core/app.dart
```diff
@@ -80,33 +78,52 @@ class _MostroAppState extends ConsumerState<MostroApp> {
...
-        debugPrint('Initial mostro deep link detected: $initialUri');
-
-        // Schedule the deep link processing after the router is ready
-        WidgetsBinding.instance.addPostFrameCallback((_) {
-          _handleInitialMostroLink(initialUri);
-        });
+      if (initialUri != null &&
+          DeepLinkInterceptor.isCustomSchemeUri(initialUri)) {
+        logger.i('Initial deep link detected: $initialUri');
+        _queueDeepLink(initialUri);
```
grunch: 🔵 **Nitpick**

**Unflagged scope widening.**

This path went from `initialUri.scheme == 'mostro'` to *any* non-web scheme, so anything `getInitialLink()` returns now reaches `DeepLinkHandler`, which answers with the red `unsupportedLinkFormat` snackbar for schemes it doesn't know.

I checked the manifests: `lightning` sits in `<queries>`, not an `intent-filter`, and iOS registers only `mostro` in `CFBundleURLSchemes` — so nothing else can actually arrive today and the risk is theoretical. Symmetry with what the router discards is the right target; just worth a line in the PR body, since it isn't mentioned.

lib/core/app_routes.dart
```diff
@@ -37,17 +37,24 @@ import 'package:mostro_mobile/features/walkthrough/providers/first_run_provider.
...
+  // go_router prefers over initialLocation; matching it asserts. Kept
+  // conditional so web still opens at the requested URL.
+  final platformDefaultLocation =
+      WidgetsBinding.instance.platformDispatcher.defaultRouteName;
+
   return GoRouter(
     navigatorKey: MostroApp.navigatorKey,
     initialLocation: '/',
+    overridePlatformDefaultLocation:
+        DeepLinkInterceptor.isCustomSchemeLocation(platformDefaultLocation),
```
grunch: 🟡 **Refactor suggestion** | _Medium_

**The discarded platform default is never used as a fallback.**

Once the override kicks in, recovering the link rests entirely on `app_links.getInitialLink()` returning the same thing. If that call fails or returns `null` on a cold start, the link is gone with no trace: it used to crash, now it's silent.

At minimum log it here; better, hand `platformDefaultLocation` to `_queueDeepLink` as a backup source so the two paths can't both come up empty.

lib/core/deep_link_interceptor.dart
```diff
@@ -62,9 +62,16 @@ class DeepLinkInterceptor extends WidgetsBindingObserver {
   }
 
   /// Check if the URI uses a custom scheme
-  bool _isCustomScheme(Uri uri) {
-    return uri.scheme == 'mostro' ||
-           (!uri.scheme.startsWith('http') && uri.scheme.isNotEmpty);
+  bool _isCustomScheme(Uri uri) => isCustomSchemeUri(uri);
```
grunch: 🟡 **Refactor suggestion** | _Medium_

**The router now depends on the interceptor for a pure predicate, and this wrapper is dead weight.**

Deduplicating the three copies is the right call — the home isn't. `app_routes.dart` importing `DeepLinkInterceptor` just for two `static bool`s inverts the natural direction: routing is the lower layer here, the interceptor is the consumer.

Extracting to `lib/core/deep_link_schemes.dart` (`isCustomSchemeUri` / `isCustomSchemeLocation`) makes all three callers symmetric and lets `_isCustomScheme` go away — with the static in scope it's now a pure alias over two call sites.

lib/core/deep_link_interceptor.dart
```diff
@@ -62,9 +62,16 @@ class DeepLinkInterceptor extends WidgetsBindingObserver {
...
+
+  /// Whether the URI uses a scheme the app resolves itself, such as `mostro:`
+  static bool isCustomSchemeUri(Uri uri) =>
+      uri.scheme.isNotEmpty && uri.scheme != 'http' && uri.scheme != 'https';
+
+  /// [isCustomSchemeUri] for an unparsed location; unparseable means no
```
grunch: 🔵 **Nitpick**

The doc comment is cut off mid-sentence — `unparseable means no` is missing its noun.

```suggestion
  /// [isCustomSchemeUri] for an unparsed location; unparseable input is
  /// treated as not custom.
```

test/core/app_deep_link_test.dart
```diff
@@ -0,0 +1,153 @@
...
+  void initialize(GoRouter router) {}
+
+  @override
+  Future<void> handleInitialDeepLink(Uri uri, GoRouter router) async {
+    if (failures > 0) {
+      failures--;
+      throw StateError('delivery failed');
+    }
+    handled.add(uri);
+  }
```
grunch: 🔵 **Nitpick** — pairs with the critical comment on `lib/core/app.dart:116`

**The double's contract differs from the collaborator it stands in for.**

`_RecordingHandler` throws; `DeepLinkHandler` never does — it catches everything and returns normally. So this is what makes the retain-on-failure test pass while the production path drops the link on every real failure.

Once the retry is driven by a return value rather than an exception, this fake should model *that* (e.g. `failures-- ; return false;`) so the test exercises the path the app actually takes.

test/core/app_deep_link_test.dart
```diff
@@ -0,0 +1,153 @@
...
+      await tester.pump();
+
+      initialLink.complete(_link);
+      await tester.idle();
+      expect(handler.handled, isEmpty);
+
+      // Any later rebuild of the app must find the link still there.
+      final container =
+          ProviderScope.containerOf(tester.element(find.byType(MostroApp)));
+      await container.read(settingsProvider.notifier).updateDefaultFiatCode('EUR');
```
grunch: 🔵 **Nitpick**

Two things here.

First, this is over 80 columns — the one genuinely new formatting violation in the PR (the rest of `dart format`'s output is the repo-wide style drift, not yours).

Second, and more interesting: driving the retry with `updateDefaultFiatCode('EUR')` couples the test to an unrelated provider, and it enshrines the behaviour I'd push back on — **a retained link fires on the next incidental rebuild**, not on an explicit retry. Once #1 is fixed and links really do get retained, that means a link can open an order minutes after it failed, in the middle of whatever the user is doing. A TTL, or a bounded explicit retry, would be better than "whichever build happens next".

test/core/app_routes_test.dart
```diff
@@ -0,0 +1,85 @@
...
+      child: Consumer(
+        builder: (context, ref, _) {
+          router = createRouter(ref);
+          return const SizedBox.shrink();
+        },
+      ),
+    ),
+  );
+  return router;
+}
```
grunch, line 44: 🔵 **Nitpick**

`createRouter(ref)` runs inside a `Consumer.builder`, which may be invoked more than once — each pass constructs another `GoRouter` that nobody disposes, and the `late router` assignment quietly depends on build ordering.

A small `StatefulWidget` creating the router in `initState`, or a plain `ProviderContainer` + `Consumer`-free call, would make this deterministic. Not blocking — the assertions themselves are good, and I confirmed the first test does fail against the unfixed router.

grunch (changes requested):
**Actionable comments posted: 11**

<details>
<summary>🧩 Walkthrough</summary>

The root-cause analysis is right and I confirmed it locally on `a0e4207d`:

- `flutter analyze lib/core test/core` → 2 pre-existing infos only (`containsSemantics` in `automation_contract_test.dart`)
- `flutter test test/core/app_deep_link_test.dart test/core/app_routes_test.dart test/core/deep_link_interceptor_test.dart` → 13/13 passing

Commits 1–2 (`overridePlatformDefaultLocation` + exact `http`/`https` matching) are solid, well-reasoned, and the `app_routes_test.dart` regression test is a real one. The findings below are all in the **delivery layer** added by commits 3 and 5.

</details>

---

## Summary

| # | Severity | Location | Issue |
|---|---|---|---|
| 1 | 🔴 Critical | `lib/core/app.dart:116-127` | Retain-on-failure never fires against the real handler — link is dropped silently |
| 2 | 🟠 High | `lib/core/app.dart:109` | `SchedulerPhase.idle` guard is narrower than Flutter's contract; a chained retry may never get a frame |
| 3 | 🟡 Medium | `lib/core/app.dart:93` | `_queueDeepLink` overwrites a pending link with no log |
| 4 | 🟡 Medium | `lib/core/app.dart:98` | Missing `mounted` guard before `ref.read` |
| 5 | 🟡 Medium | `lib/core/app_routes.dart:47-54` | Discarded platform default is not used as a fallback source |
| 6 | 🟡 Medium | `lib/core/deep_link_interceptor.dart:65` | Router→interceptor dependency for a pure predicate; redundant private wrapper |
| 7 | 🔵 Low | `lib/core/deep_link_interceptor.dart:71` | Truncated doc comment |
| 8 | 🔵 Low | `lib/core/app.dart:84` | Unflagged scope widening from `mostro:` to any custom scheme |
| 9 | 🔵 Low | `test/core/app_deep_link_test.dart:33` | Fake's contract differs from `DeepLinkHandler` (pairs with #1) |
| 10 | 🔵 Low | `test/core/app_deep_link_test.dart:127` | Retry driven by an unrelated rebuild; >80 col |
| 11 | 🔵 Low | `test/core/app_routes_test.dart:25` | `createRouter` inside a `Consumer.builder` |

## Verdict

**Request changes** on #1: the guarantee the last commit claims — *"The pending link now also survives a failed delivery"* — does not hold against `DeepLinkHandler`, which swallows every ordinary failure and returns normally. The test that backs it uses a double with a different contract, so it passes while the production path drops the link. #2 is a one-liner worth folding into the same push.

If you'd rather unblock the crash fix now, commits 1–2 stand on their own and could merge separately; as it stands commits 3–5 add a safety net that catches nothing.

## Test coverage gaps

Not blocking, but the cases most likely to break are the ones not covered: a second link arriving while one is pending (#3), re-entrancy of `_deliverPendingDeepLink` (#2), and a custom-scheme platform default when `app_links` returns nothing (#5).

## Note on formatting

`dart format` rewrites all four source files, but the whole repo is on the previous formatter style and there is no format gate in `.github/workflows/`, so this is not on you — except the one new >80 col line flagged inline.

21Mill:
You were right about #1, and I took the exit you offered: this PR is now the crash fix only, and the delivery layer moved to #692.

### On #1

The retain-on-failure guarantee was false, and the reason it looked true is the one you named: my double threw, `DeepLinkHandler` doesn't. I traced the five return paths you listed and they all return normally, so the `catch` was unreachable in production. A green test over a double whose contract differs from the collaborator is not evidence, and I shipped it as if it were.

The fix in #692 is not "retain on any failure" either — that would reopen an order the user already dismissed, which is your #10. `handleInitialDeepLink` returns `Future<bool>` and only *"the app could not attempt it"* (no navigator, or another link already being opened) is retained; anything the user already got an answer for counts as done. Retries ride on consecutive frames and stop after ten.

Two things fell out of writing that: the 2 s duplicate window was stamped **before** the context check, so a link that was never attempted came back and was taken for a duplicate of itself; and the same early return left the loading dialog up.

### Split

- **This PR**: commits 1–2, plus #6 (the predicate moved to `lib/core/deep_link_schemes.dart`, so `app_routes.dart` no longer imports the interceptor and the `_isCustomScheme` alias is gone), #7 (the truncated doc), the log for the discarded platform default from #5, and #11 (the router test holds one router and disposes it).
- **#692**: #1, #2 (the `== idle` guard is gone — `ensureVisualUpdate` is already a no-op in the other phases, as you quoted), #3, #4, #5's real fallback, #8, #9 and #10.

I also wrote into the body of this PR what the split costs: with only these commits the crash is gone, but on a slow cold start the link can still fail to open the order, since delivery is still `main`'s 100 ms callback. On the device that window is ~0.9 s, so it is not theoretical. #692 closes it.

### Not addressed

#2 has no test. I could not build one that distinguishes the phases — the navigator mounts in the same frame as the router — so it rests on the source contract you quoted rather than on a red-to-green.

### Verification

`flutter analyze lib/core test/core` clean but for the two pre-existing `containsSemantics` infos; `flutter test test/core/` green; full `flutter test` 897 passing with the same 11 pre-existing load failures (#606). On the OnePlus 8T: cold start opens the order with no assertion, foreground still works, plain launch shows the order book.

PR #694, opened 2026-08-30
perf: value equality for Order/PaymentFailed and no-op state guards

grunch:
## Summary

Item **2.3** of the performance plan. `OrderState.==` compares its `order` and `paymentFailed` payloads (`order_state.dart:53-65`), but both classes had identity equality only. Every replayed history message and every duplicate therefore produced a "new" state: N notifier emissions at startup and per incoming event, each one recomputing `filteredTradesWithOrderStateProvider` (full book sort), flashing dispute FutureProviders back to loading, and rebuilding every `MostroReactiveButton`/`TradeDetailScreen` watcher.

## Changes

- `Order` and `PaymentFailed` extend `Equatable` (all fields in `props`; `const` constructors preserved).
- `OrderNotifier.sync()` and `updateStateFromMessage()` guard with `if (next != state)` so identical replays no longer notify watchers.
- `order_state_test.dart` had the old behaviour **pinned on purpose** ("pinned here so a fix shows up as a deliberate change"); that pin is flipped to the fixed contract, per its own comment.

Note: `Session` immutability (also mentioned in plan item 2.3) is deliberately left out — `Session` has live setters used across the codebase and deserves its own focused PR.

## Test plan

- [x] New `test/models/order_equality_test.dart` (RED on `main`): Order/PaymentFailed value equality; `OrderState`s built from equal payloads compare equal
- [x] Updated pinned test in `order_state_test.dart`
- [x] Full `flutter test` — all green
- [ ] Manual: cold start with many trades — My Trades paints once per real change; order flows unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate order-state notifications when identical updates are received or replayed.
  * Ensured meaningful order changes continue to update the interface correctly.
  * Improved equality handling for orders, payment failures, and order states.

* **Tests**
  * Added coverage for value equality and notification behavior.
  * Updated expectations for equivalent order states.


grunch:
## Review

The equality work itself is correct. I verified the parts that could silently break:

- `Order.props` covers all 17 fields, all `final`; no `Order` subclass exists, so Equatable's `runtimeType` check is a non-issue.
- No call site depends on `Order` identity (no `Set<Order>`, no `Map<Order, _>`, no `identical`/`==` used to mean "same instance"), so the switch to value semantics can't change behaviour anywhere.
- Every field of `OrderState` is in its `==`, and every nested payload it compares (`PaymentRequest`, `CantDo`, `Dispute`, `Peer`, `UserInfo`) already had value equality — so no real change can be swallowed by a `!=` guard.
- `MostroReactiveButton` resolves its spinner from `mostroMessageStreamProvider`, not from state changes, so suppressing duplicate emissions doesn't strand a button in loading.
- Equatable changes `Order.toString()` in debug, but nothing logs an `Order`/`OrderState` payload (`order_state.dart:398` logs only status/action), so no new output.

### One real gap: the guards miss the path the PR is about

The description targets "N notifier emissions at startup **and per incoming event**". The startup half lands; the per-event half doesn't.

`AbstractMostroNotifier` is a `StateNotifier`, whose default `updateShouldNotify` is `!identical(old, new)`. A new-but-equal instance still fans out. This hierarchy assigns `state` in **13** places; the PR guards **2**, and the live path is not among them:

- `abstract_mostro_notifier.dart:123` — `state = state.updateWith(msg)`, run for every gift wrap the subscription receives, duplicates included. The `_processedEventIds` dedupe at `handleEvent` (line ~197) only guards the *side effects*; the assignment above it happens first, every time. Relays re-deliver freely, so this is the common case, not the edge one.
- `abstract_mostro_notifier.dart:351/425/545/634` — `copyWith(peer:)` / `copyWith(dispute:)` refreshes that frequently carry identical data.
- `add_order_notifier.dart:74/95/127`.

The repo already has the idiomatic fix one PR back: `SettingsNotifier` overrides `updateShouldNotify` (#689). Doing the same in `AbstractMostroNotifier` covers all 13 sites with one mechanism and makes the two call-site guards redundant.

### CodeRabbit's nitpick

Agreed — `test/data/models/` mirrors `lib/data/models/` and is where the sibling payload tests live (`payload_models_test.dart`, `protocol_payloads_test.dart`).

### Changes pushed

Branch [`claude/review-pr-694-hzzri5`](https://github.com/MostroP2P/mobile/compare/perf/order-value-equality...claude/review-pr-694-hzzri5) (commit `12fc2f6`, on top of `b279c42`):

1. `updateShouldNotify(old, current) => old != current` in `AbstractMostroNotifier`; the two ad-hoc guards dropped as redundant.
2. `test/models/order_equality_test.dart` → `test/data/models/`.
3. `test/features/order/notifiers/order_state_no_op_emission_test.dart` — drives a real message twice through the live subscription and asserts one emission, plus a companion asserting a changed field still emits. Both are RED without the override.

`flutter analyze` clean; `flutter test` 1195/1195 green (1193 on this PR's head + the 2 new).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #700, opened 2026-08-30
perf: cache key derivation and make session ECDH lazy

grunch:
## Summary

Item **3.1** of the performance plan (crypto costs on the UI isolate — all of it pure-Dart BigInt secp256k1, 5–30 ms per scalar multiplication on a mid phone).

Decoding one session re-ran the full derivation pipeline every time: `BIP32.fromBase58` (base58check) + 5 CKD levels + an EC multiplication for the pubkey (`KeyManager.deriveTradeKeyPair`), plus 1–2 ECDH computations in the `Session` constructor (`computeSharedKey` also wraps the secret in a `NostrKeyPairs`, another multiplication). This ran **per session** at startup and again from the **30-minute cleanup**, which re-read and re-decoded every session from disk although `SessionNotifier` already held them in memory. ≈3–5 multiplications × N sessions ≈ 0.5–2 s of synchronous UI-thread work, repeated every half hour.

## Changes

- **`KeyDerivator`**: parsed BIP32 root cached per master key; derived child hex memoized per (root, index) with a bounded map. Deterministic, so behaviour pinned against the NIP-06 vectors.
- **`KeyManager`**: trade key pairs memoized per index (`identical` pinned by test), cleared on master-key change; `deriveTradeKey`/`deriveTradeKeyFromIndex` reuse the cached master hex instead of a secure-storage read per call. `@visibleForTesting debugSetMasterKey` allows storage-free tests.
- **`Session`**: `sharedKey`/`adminSharedKey` computed lazily on first access; the constructor and the `peer`/`setAdminPeer` setters only record the inputs (the admin pubkey validation stays in the constructor path).
- **`SessionNotifier._cleanup`**: iterates the in-memory sessions instead of `getAllSessions()` (which re-decoded — and before this PR re-derived — every row). Note: sessions written by other isolates between cleanups are picked up at next app start, same as before for anything the notifier didn't know about.

## Test plan

- [x] New `test/features/key_manager/key_manager_cache_test.dart` (RED on `main`): per-index memoization by identity, distinct indices, master-key change invalidation, derivator cache correctness against NIP-06 vectors
- [x] `key_derivator_test.dart` (NIP-06 vectors) — green
- [x] Subscriptions + chat suites (consumers of `sharedKey`/`adminSharedKey`) — green
- [x] Full `flutter test` — all green
- [x] `flutter analyze` — no new issues
- [ ] Manual: cold start with many trades — no multi-second freeze before the list paints; chat and dispute filters still built correctly

[...]

PR #703, opened 2026-08-31
perf: run media sanitizing and file crypto off the main isolate

grunch:
## Summary

Item **3.5** of the performance plan. Three CPU-heavy media paths ran on the UI isolate:

- **Image sanitizing** (`package:image`, pure Dart): decode + re-encode of a 12 MP photo takes 1–5 s, per image sent (`media_validation_service.dart`).
- **Whole-file ChaCha20-Poly1305** (pointycastle, files up to 25 MB): per media send, per download, and per history auto-download (`encryption_service.dart`).
- **Macro-pattern scan** (O(n·m) over the whole file) in document validation.

## Changes

- `validateAndSanitizeImage` / `validateAndSanitizeImageLight` keep their public signatures but execute through `Isolate.run` (byte buffers and `MediaValidationResult`/exceptions transfer across the boundary; propagation pinned by test).
- `EncryptionService.encryptToBlobAsync` / `decryptFromBlobAsync`: `Isolate.run` over the existing sync implementations (which stay for callers that need them and for the isolate itself). The encrypted image/file upload services now use the async variants in both directions.
- The macro scan runs through `Isolate.run` at both call sites.

**Left for 3.4 (worker isolate):** per-message NIP-44/Schnorr work — batching those into a long-lived worker is the remaining phase-3 item, done after the caches (#700, #701, #702) so the isolate doesn't inherit redundant work.

## Test plan

- [x] New `test/services/media_crypto_isolate_test.dart` (RED on `main`): async blob roundtrip + cross-compat with sync, tampered-blob failure across the isolate, PNG light/heavy sanitize results, garbage-input error propagation
- [x] `file_messaging_test.dart` + chat suites — green
- [x] Full `flutter test` — all green
- [ ] Manual: send a large photo in chat — UI stays responsive during sanitize+encrypt; open received media — decrypts correctly

### DIFF
--- a/lib/services/encrypted_file_upload_service.dart
+++ b/lib/services/encrypted_file_upload_service.dart
@@ -80,17 +80,12 @@ class EncryptedFileUploadService {
       );
       
       // 2. Encrypt with ChaCha20-Poly1305
-      final encryptionResult = EncryptionService.encryptChaCha20Poly1305(
+      final encryptedBlob = await EncryptionService.encryptToBlobAsync(
         key: sharedKey,
         plaintext: validationResult.validatedData,
       );
-      
-      final encryptedBlob = encryptionResult.toBlob();
       logger.i(
         '🔐 File encrypted successfully: ${encryptedBlob.length} bytes '
-        '(nonce: ${encryptionResult.nonce.length}B, '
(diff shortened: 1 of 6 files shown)


Catrya (approved):
tACK

PR #710, opened 2026-08-31
test: fix the duplicate-envelope race pin for cross-isolate unwrapping

grunch:
## Summary

`dispute_chat_duplicate_envelope_test.dart` fails on `main`. Root cause: the pin was added (during #702's review) while envelope unwrapping was synchronous; #705 later moved the per-message crypto into `Isolate.run`, and `pumpEventQueue` cannot observe cross-isolate port replies — the assertion runs before the valid envelope finishes verifying. The two PRs crossed in flight (#705 was stacked on #701, whose branch predated this test), so CI never ran the combination.

**The product behaviour is correct** (verified by log trace): the forged copy fails the p-tag check and the valid copy is processed on its own; only the test's wait was implementation-timed.

## Changes

- Bounded 5 s condition wait on the notifier's messages (same pattern already applied to `dispute_chat_reload_test` in #705).

## Test plan

- [x] `test/features/disputes/` — all green (previously 1 failing on `main`)

### DIFF
--- a/test/features/disputes/dispute_chat_duplicate_envelope_test.dart
+++ b/test/features/disputes/dispute_chat_duplicate_envelope_test.dart
@@ -174,7 +174,17 @@ void main() {
 
     nostrService.controller.add(forged);
     nostrService.controller.add(real);
-    await pumpEventQueue(times: 200);
+    // Unwrapping runs through Isolate.run: wait on the observable condition
+    // (bounded) instead of a microtask pump count, which cannot see cross-
+    // isolate port replies.
+    final deadline = DateTime.now().add(const Duration(seconds: 5));
+    while (container
+            .read(disputeChatNotifierProvider(disputeId))
+            .messages
+            .isEmpty &&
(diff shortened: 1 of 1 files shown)


Catrya (approved):
ACK

grunch:
Superseded by #709, closing.

The same fix landed on `main` in 0381ff85 (#709): `pumpEventQueue(times: 200)` in `dispute_chat_duplicate_envelope_test.dart` is replaced by a bounded poll on the observable condition, since unwrapping now runs through `Isolate.run` and a microtask pump count cannot see cross-isolate port replies. Only the shape differs — a `_waitFor` helper there vs. the inline loop here.

This PR's diff no longer applies (the `pumpEventQueue(times: 200)` line it removes is gone from `main`), so leaving it open would only produce a conflict.

All the PRs that were red on this test (#711–#715, #717, #718, plus #716 through its base) have had `main` merged in and are picking up the fix from there.

PR #713, opened 2026-08-31
perf: cache sessions and chat keys in the background isolate

grunch:
## Summary

Item **4.4** of the performance plan. For **every incoming event** the background notification pipeline: opened `mostro.db`, constructed and `init()`-ed a `KeyManager` (secure-storage reads), decoded every session (trade-key derivation) and derived every session's `ChatKeys` — **twice per event** (`_decryptAndProcessEvent` + `showLocalNotification` both loaded sessions), plus again inside the peer-update path. Pure CPU/battery drain while the app sleeps, worst under message bursts.

## Changes

- **`BackgroundSessionCache`** (new, pure class): session list + node pubkey memoized behind a 2-minute TTL; a concurrent burst shares one in-flight load; `ChatKeys` derived once per shared key (bounded map).
- The pipeline's read sites go through the cache; `_maybeUpdateSessionWithPeer` **invalidates** it right after its `putSession`, so the next event sees the fresh peer. A fresh service start naturally begins empty.
- Foreground/background DB co-access is unchanged by this PR (reads just stop repeating). Serializing cross-isolate writes remains the known follow-up from the plan.

## Test plan

- [x] New `background_session_cache_test.dart` (compile-RED first): single load per TTL window, invalidate reloads, TTL expiry, ChatKeys identity-memoized, concurrent burst shares one load
- [x] Notifications suite 48/48; full suite in halves — green except `dispute_chat_duplicate_envelope_test.dart`, which is the known pre-existing failure fixed by #710 (not in this branch's base)
- [ ] Manual: background the app, receive several trade/chat events — notifications correct; take an order while backgrounded — peer chat notification works after the update

## Summary by CodeRabbit

- **Improvements**
  - Background notifications now process more efficiently by reusing session and chat information during short periods of activity.
  - Reduced repeated database access and cryptographic setup, helping notification handling respond more quickly.
  - Cached information is refreshed automatically when session data changes or becomes outdated, ensuring notifications continue using current details.

### DIFF
--- a/lib/features/notifications/services/background_notification_service.dart
+++ b/lib/features/notifications/services/background_notification_service.dart
@@ -1,3 +1,4 @@
+import 'package:mostro_mobile/features/notifications/services/background_session_cache.dart';
 import 'dart:convert';
 import 'dart:math';
 
@@ -168,7 +169,7 @@ Future<void> showLocalNotification(NostrEvent event) async {
     final mostroMessage = await _decryptAndProcessEvent(event);
     if (mostroMessage == null) return;
 
-    final sessions = await _loadSessionsFromDatabase();
+    final sessions = await backgroundSessionCache.sessions(_loadSessionsFromDatabase);
     final matchingSession = sessions.cast<Session?>().firstWhere(
       (session) => session?.orderId == mostroMessage.id,
       orElse: () => null,
@@ -252,7 +253,7 @@ Future<MostroMessage?> _decryptAndProcessEvent(NostrEvent event) async {
       return null;
     }
 
-    final sessions = await _loadSessionsFromDatabase();
+    final sessions = await backgroundSessionCache.sessions(_loadSessionsFromDatabase);
 
     // Chat (kind 14 envelope): the outer author is the K_sign key derived
     // from the conversation's shared key, so match by author, not recipient.
@@ -262,7 +263,7 @@ Future<MostroMessage?> _decryptAndProcessEvent(NostrEvent event) async {
       for (final session in sessions) {
         final adminShared = session.adminSharedKey;
         if (adminShared == null) continue;
-        final chatKeys = ChatKeys.fromSharedKey(adminShared);
+        final chatKeys = backgroundSessionCache.chatKeysFor(adminShared);
         if (event.pubkey == chatKeys.sign.public) {
           return _processAdminDm(event, session, chatKeys);
         }
@@ -270,7 +271,7 @@ Future<MostroMessage?> _decryptAndProcessEvent(NostrEvent event) async {
       for (final session in sessions) {
         final shared = session.sharedKey;
         if (shared == null) continue;
-        final chatKeys = ChatKeys.fromSharedKey(shared);
+        final chatKeys = backgroundSessionCache.chatKeysFor(shared);
         if (event.pubkey == chatKeys.sign.public) {
           return _processPeerChat(event, session, chatKeys);
         }
@@ -343,7 +344,8 @@ Future<MostroMessage?> _handleTradeKeyEvent(NostrEvent event, Session session) a
   // decrypts straight to the tuple. Both converge on jsonDecode below.
   final String? content;
   if (event.kind == 14) {
-    final mostroPubkey = await _loadMostroPubkey();
+    final mostroPubkey =
+        await backgroundSessionCache.mostroPubkey(_loadMostroPubkey);
     if (mostroPubkey == null) {
       logger.w('No Mostro pubkey available, cannot decrypt kind-14 event');
       return null;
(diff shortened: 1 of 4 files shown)


Catrya (changes requested):
Reviewed with the branch applied locally: `flutter analyze` clean,
`test/features/notifications/` 48/48, full suite 1295 pass / 1 fail (the known
`dispute_chat_duplicate_envelope_test`, fixed by #710 — that's also what the red
CI here is, so #710 should land first and this should rebase).

The premise checks out: the service stops on foreground
(`mobile_background_service.dart:120` → `stopSelf()` at `background.dart:355`),
so the isolate never coexists with the foreground, and there is exactly one
session write site in the isolate (`background_notification_service.dart:455`),
which does invalidate. The saving is real and worth having.

Two correctness issues before merge, though — both produce the same symptom
(silently dropped notifications for up to the full 2-minute TTL).

### 1. `invalidate()` is lost when a load is already in flight

`invalidate()` clears `_sessions`/`_loadedAt` but not `_inFlight`. A load that
started before the invalidation completes afterwards and writes its
**pre-write** snapshot back into `_sessions` with a fresh `_loadedAt`. The
invalidation is swallowed and the stale list is served for the whole TTL.

Reachable in production: the three subscription sites use
`subscription.listen((event) async { ... })` (`background.dart:153, 233, 321`),
which does not await the callback, so events process concurrently — and
`_loadSessionsFromDatabase` is slow enough (db open + secure storage + key
derivation) to leave a wide window:

1. event B calls `sessions()` → load L starts
2. event A does `putSession` for the peer → `invalidate()`
3. L resolves with the snapshot that has no peer, and caches it

For the next 2 minutes the peer-chat branch of `_decryptAndProcessEvent` finds
no `session.sharedKey` match, so peer chat notifications are dropped. That is
exactly the unchecked manual-test item in the description ("take an order while
backgrounded").

A generation counter fixes it — verified locally, the repro below goes green and
all five new tests still pass:

```dart
int _generation = 0;

// in sessions(), before creating the load:
final generation = _generation;
final load = () async {
  try {
    final loaded = await loader();
    if (generation == _generation) {
      _sessions = loaded;
      _loadedAt = DateTime.now();
    }
    return loaded;
  } finally {
    if (generation == _generation) _inFlight = null;
  }
}();

// same guard around the await in mostroPubkey()

void invalidate() {
  _generation++;
  _inFlight = null;
  // ...
}

Repro:

final inFlight = cache.sessions(loader);          // load starts
await Future<void>.delayed(const Duration(milliseconds: 10));
cache.invalidate();                                // concurrent local write
await inFlight;                                    // stale snapshot lands
expect((await cache.sessions(loader)).single.orderId, 'o2'); // gets 'o1'
```

2. Failed loads are cached (agrees with the Codex P1)

_loadSessionsFromDatabase catches everything and returns [], which the cache
stores as a successful load. Before this PR each event retried the database
independently; now one transient failure is pinned for two minutes and every
event in that window is discarded. Not hypothetical — this PR's own description
notes foreground/background db co-access is still unserialized, and
FlutterSecureStorage reads can fail in the background on a locked device.

Smallest fix is to let _loadSessionsFromDatabase rethrow instead of returning
[]. I checked that it's safe: both call sites are already inside try/catch
(showLocalNotification:167, _decryptAndProcessEvent:245), sessions()
already doesn't memoize when the loader throws, and retryNotification retries
three times with backoff — strictly better than today. Worth keeping the
distinction explicit, though: caching a genuinely empty list (user has no active
trades) is correct and desirable.

On the other bot comments

- CodeRabbit, "cache a missing Mostro pubkey for the TTL" — I'd decline
  this. _loadMostroPubkey also returns null on exception
  (background_notification_service.dart:536), so caching null would turn a
  transient SharedPreferences failure into two minutes of undecryptable kind-14
  events. The current behaviour is accidentally the right one. Would only be
  safe if failure were distinguished from a genuinely absent value.
- Codex, "add integration coverage" — fair by the letter of AGENTS.md:24,
  but integration_test/ currently holds three order-creation tests and no
  harness at all for driving the background isolate. Building that is its own
  PR; I wouldn't block on it here.

Nits

- The new import landed on line 1, above dart:convert. flutter_lints
  doesn't enable directives_ordering so analyze stays clean, but still.
- sessions() hands out the live internal list; List.unmodifiable would stop
  a future call site from mutating the cache by accident.
- mostroPubkey() doesn't share an in-flight load the way sessions() does, so
  a burst does N SharedPreferences reads. Cheap, but inconsistent.
- final load = () async {...}(); reads better as a named local function.
- The update-settings handler (background.dart:292) doesn't invalidate.
  Low risk since the node can't be switched while backgrounded, but it's one
  line.

Issue #611, opened 2026-06-10
Align CI Flutter version with current stable to resolve the analyzer 7→8 toolchain split


Catrya:
## Problem

  The repository's committed `pubspec.lock` and CI are pinned to a Flutter version that sits on the **analyzer 7.x** side of a breaking toolchain boundary, while current Flutter stable is on **analyzer 8.x**. A single committed lockfile cannot serve both sides cleanly, which produces a developer-experience break for any contributor on current Flutter.

  Concretely:

  - **CI runs on Flutter 3.32.2** (`.github/workflows`, `FLUTTER_VERSION: '3.32.2'`) → Dart 3.8 → only `analyzer 7.x` is installable (`analyzer 8.x` requires `sdk: ^3.9.0`).
  - The committed `pubspec.lock` pins `analyzer 7.7.1` + `source_gen 3.1.0` — a self-consistent, working combination on CI. This is why CI and releases are green.
  - **Contributors on Flutter 3.41+/3.44** (Dart 3.11+) get `analyzer 8.x` on a clean `flutter pub get`, but the lock keeps `source_gen 3.1.0` (which declares support for `analyzer <9.0.0` even though its code calls
  `DartObjectImpl.getInvocation()`, a method removed in analyzer 8.x). Result: `dart run build_runner build -d` fails with:

    source_gen-3.1.0/lib/src/constants/revive.dart:82:40: Error: The method
    'getInvocation' isn't defined for the type 'DartObjectImpl'.

  This is the root cause behind #606. It is not a project code bug and does not affect the app, the build output, or CI (generated `*.g.dart`/`*.mocks.dart` are not committed and CI regenerates them on 3.32.2). It is a toolchain-version split that blocks local code generation for anyone on current Flutter stable.

  #### Why this is its own issue

  The same analyzer 7→8 wall surfaces through more than one dependency chain:

  - **#606** — `source_gen` (pulled via `mockito` → `build_runner`). This is the chain that actually breaks in this repo.
  - **#512** — the Riverpod ecosystem pins `analyzer ^7.0.0`. Note: this repo currently uses **no Riverpod code generation** (no `@riverpod` / `riverpod_annotation`; `riverpod_generator`/`custom_lint`/`riverpod_lint` are not in the dependency tree), so the Riverpod pin is not an active blocker here, but it is part of the same modernization frontier.

[...]

PR #717, opened 2026-09-01
perf: lazy notifiers for terminal orders and a bounded media cache

grunch:
## Summary

Item **5.4** of the performance plan. Startup eagerly created an `OrderNotifier` (storage watcher + order-book listener) and a `ChatRoomNotifier` (full history decrypt) for **every** session in the retention window — finished trades included; with the default 30-day window, N was "every trade of the month", none ever disposed. Each chat notifier also held its decrypted media bytes until process exit.

## Changes

- **Lazy terminal orders**: app init skips sessions whose order's last stored message reports a terminal status (`Status.isTerminal` via the new pure `isTerminalOrderMessage`; a missing message or non-order payload stays eager — conservative). The non-autoDispose families still build those notifiers **on demand** the moment a screen watches them (trade detail, status filtering, chat history), so nothing is lost — it just stops costing at startup and per event.
- **Bounded media cache**: `MediaCacheMixin` gets a combined 32 MB budget with LRU eviction (oldest first); evicted media re-decrypts on demand from the stored blob. `clearMediaCaches` resets the accounting.

## Test plan

- [x] New `lazy_terminal_and_media_cache_test.dart` (compile-RED first): terminal/live/ambiguous decision table; LRU eviction under the byte cap
- [x] Chat + order suites — green
- [x] Full suite halves — features shows only the known #710 failure; rest all green
- [ ] Manual: account with many finished trades — startup no longer builds their notifiers (log volume), opening an old trade detail still works; long chat with many photos stays under budget

## Summary by CodeRabbit

* **New Features**
  * Added a shared 32 MiB cache for decrypted images and files, automatically removing least-recently-used items when capacity is reached.
  * Improved app startup by deferring initialization for older, settled orders while keeping active, recent, or ambiguous conversations immediately available.
  * Ensured encrypted images reload automatically after cache eviction.

* **Bug Fixes**
  * Corrected media cache accounting across conversations and separate media types.
  * Prevented oversized media entries from being immediately removed.
  * Improved handling of invalid and varied message timestamps during startup.


Catrya (changes requested):
## Request changes

`flutter analyze` is clean and the `shared/`, `chat/` and `order/` suites are
fully green here (444 tests). The concern is the premise of the lazy-init half,
not the tests.

### Blocking 1: `isTerminal` is a cleanup predicate, not "nothing else will arrive"

`Status.isTerminal` is documented as the set of states "where the session can be
safely deleted **during cleanup**" (`status.dart:31-32`), and its only current
caller is `session_notifier.dart:104-113`, which applies it exclusively to
sessions already **past** the expiration cutoff. Reusing it inside the retention
window silently changes its meaning to "no further message is expected", and that
is not true for at least three flows:

- **`settledHoldInvoice`** — documented eight lines below in the same file
  (`status.dart:50-52`) as "the window between the release and the payout, where
  the buyer may want to replace a wrong invoice". If the app starts during that
  window the notifier is skipped, so an incoming `payment_failed` / add-invoice
  request produces no notification and no state update.
- **`canceled` / `canceledByAdmin` / `settledByAdmin`** — `bond-slashed` is a
  trailing notice that arrives *after* the terminal status. The deferred 60s
  session deletion (`abstract_mostro_notifier.dart:32-34`) exists precisely so it
  can still be received. With the notifier skipped, the forfeiture notice is
  silent.
- **`success`** — the rate / rate-received exchange.

`MostroService._onData` still persists these messages, so nothing is lost on
disk — what is lost is the live reaction (notification, navigation, session
cleanup), since order notifications are only emitted from
`AbstractMostroNotifier.handleEvent` and the background service suppresses itself
while the app is in the foreground.

The new test covers `canceled`, `success` and `canceledByAdmin` but not
`settledHoldInvoice`, which is the one that matters most here.

Suggestion: don't reuse `isTerminal`. Define the predicate this call site actually
needs — "finished *and* nothing trailing is expected" — excluding at minimum
`settledHoldInvoice`, and consider a short grace window after the terminal message
so trailing notices (`bond-slashed`, rate) still land eagerly.

### Blocking 2: terminal orders also lose their eager `ChatRoomNotifier`

The `continue` skips `ref.read(chatRoomsProvider(session.orderId!))` too. Those
notifiers are the only consumers of `SubscriptionManager.chat`, which is a
`StreamController.broadcast()` — it drops events while it has no listener. So a
P2P message on a finished trade ("received, thanks") that arrives before the user
opens the Chats tab is discarded: not displayed, not written to disk, and the
cursor does not advance. `ChatRoomsNotifier` only builds those notifiers when the
chat list renders (`_chatsForSessions`), which is lazy.

Keeping the chat notifier eager while skipping only the order notifier would avoid
this, at the cost of the history decrypt — or the chat side needs the same
listener-gap fix discussed in #714.

### Non-blocking

- **The media budget is per notifier, not global.** `mediaCacheMaxBytes` is
  `static const` but `_mediaBytes` is an instance field, so the ceiling is 32 MB ×
  number of conversations. Given the stated motivation ("these notifiers live for
  the whole app run"), either account globally or pick a much smaller
  per-conversation budget.
- **Evicted metadata is retained.** `_mediaTouch` drops the bytes from
  `_imageCache`/`_fileCache` but never from `_imageMetadata`/`_fileMetadata`. For
  files this is harmless (`encrypted_file_message.dart:70-83` re-parses metadata
  and re-downloads on a cache miss), but `encrypted_image_message.dart` only calls
  `_loadImage()` from the `initState` post-frame callback: a still-mounted widget
  whose bytes were evicted renders `_buildLoadingWidget()` indefinitely, since
  `build` sees `cachedImage == null`, `_isLoading == false`, `_errorMessage ==
  null`. Narrow (needs >32 MB in one conversation) but it is a new failure mode
  introduced by eviction. Simplest fix: re-trigger the load on a cache miss in
  `build`.
- **It is FIFO, not LRU.** `_mediaLru` is only touched on insert; `getCachedImage`
  / `getCachedFile` don't promote. Fine in practice, but the comment and the field
  name claim otherwise.
- `debugMediaCacheBytes` is public production API — worth `@visibleForTesting`.
- The three new imports were added above `import 'dart:async';` in
  `app_init_provider.dart`, breaking the usual dart:/package: ordering.
- Worth noting for expectations: `trades_list_item.dart:33` watches
  `orderNotifierProvider` per row, so the saving is a startup-only one that
  unwinds as soon as the user opens My Trades. Still worth having — just not a
  permanent N reduction.

grunch:
Feedback addressed in e26100f.

## Blocking

**Catrya 1 / Codex P1 — `isTerminal` is a cleanup predicate, not "nothing else will arrive".** You are right, and reusing it was the wrong call. Replaced with `isSettledOrderMessage`, a predicate defined for this call site, plus a 24h trailing-notice window on the last message. Eager, explicitly:
- `settledHoldInvoice` — the release-to-payout window where the buyer may still replace a wrong invoice;
- `canceled` — `OrderNotifier.sync()` calls `reconcileCanceledBondedSession()` on this status, which is exactly the restart path Codex flagged (app killed inside the 60s grace window); `bond-slashed` also trails it;
- `success` — the rating exchange has no time bound, so a window would not make it safe.

That leaves `canceledByAdmin`, `settledByAdmin`, `completedByAdmin`, `cooperativelyCanceled` and `expired`, and only after 24h so `bond-slashed` and ratings still land eagerly. Timestamps are unit-normalized first (the daemon sends seconds), and an unknown age counts as live. Tests cover each excluded status, the grace window, and the seconds unit.

**Catrya 2 — terminal orders lost their eager `ChatRoomNotifier`.** Fixed: the `continue` is gone, and the chat notifier is now created for every session with a peer regardless of order status. Only the order notifier is made lazy. `SubscriptionManager.chat` being a broadcast stream that drops events with no listener makes this non-negotiable; the history decrypt is the price.

## Non-blocking

- **Per-notifier budget.** Now global: a static LRU of `(owner, messageId, bytes)` with a shared byte total, so 32 MB is a real ceiling rather than 32 MB × conversations. `clearMediaCaches()` releases only that owner's entries. Test asserts two conversations do not double it.
- **FIFO, not LRU (Codex P2 / CodeRabbit).** `getCachedImage` / `getCachedFile` now promote on a hit without touching byte accounting. Test reads an older entry before pushing past the limit and asserts it survives.
- **Evicted bytes with retained metadata.** `encrypted_image_message.dart` now re-requests the load from `build` on a cache miss, so a still-mounted widget cannot sit on `_buildLoadingWidget()` forever. Metadata is deliberately kept on eviction — it is small and the reload re-caches the blob.
- `debugMediaCacheBytes` is `@visibleForTesting`.
- Imports reordered (`dart:` first), and the `if(` guard is `dart format` clean (CodeRabbit).
- Tests split into `test/shared/providers/app_init_provider_test.dart` and `test/shared/mixins/media_cache_mixin_test.dart` to mirror their source modules (CodeRabbit nitpick).
- **On expectations:** agreed on `trades_list_item.dart:33` — this is a startup-only saving that unwinds when My Trades opens. With the narrower predicate the win is smaller still; the eviction bound and the chat-notifier correctness are the durable parts.

`flutter analyze` clean (2 pre-existing `containsSemantics` infos in `test/core/automation/automation_contract_test.dart`). Full suite 1301/1301 green, 11 tests in the two new files.

Catrya (changes requested):
### 1. An entry larger than the budget evicts itself, and the new post-frame callback turns that into an infinite loop

Reproduced:

```
insert a 32 MB + 1 byte image
  bytes tracked       = 0
  cached after insert = false
```

`_evict()` runs `while (_totalBytes > max && _lru.isNotEmpty)`, so it empties the whole list — including the entry that was just inserted.

The PR then added this inside `build()`, at `encrypted_image_message.dart:95`:

```dart
WidgetsBinding.instance.addPostFrameCallback((_) => _loadImageIfNeeded());
```

which closes the cycle: `build` → not cached, not loading, no error → post-frame → `_loadImage()` → download + decrypt → cache → self-evict → `setState(_isLoading = false)` → `build` → again. **Downloading and decrypting in a loop for as long as the widget stays mounted.**

I looked for a size cap in `encrypted_image_upload_service.dart`, `encrypted_file_upload_service.dart` and across `features/chat/` — there is **none**; the only hits are "MB" formatters. So a >32 MB image is reachable: a large PNG, a panorama, or a peer on a modified client.

One-word fix: stop eviction before the list empties —

```dart
while (_totalBytes > mediaCacheMaxBytes && _lru.length > 1) {
```

The new entry survives, the budget is only exceeded in that pathological case, and the loop disappears. Worth a test with an entry larger than the cap; there isn't one today.

There's a more benign variant I did not prove: several mounted images summing over 32 MB could take turns evicting each other and trigger the same re-load cycle. Flagging it as a suspicion, not a fact.

### 2. `pubspec.lock` doesn't belong in this PR

44 lines bumping `analyzer` 7.7.1 → **8.4.1**, `_fe_analyzer_shared` 85 → 91, `test` 1.26.2 → 1.30.0, plus `meta`, `material_color_utilities`, `dart_style`, `characters`. None of it relates to lazy notifiers or the media cache.

Two things confirm it's accidental:

- **It's the only PR in the series that touches it.** #711, #712, #713, #714 and #718 leave `pubspec.lock` alone.
- **It doesn't survive.** Running the tests triggered an implicit `flutter pub get` that reverted the whole lock to `main`'s values — 22 lines back, `8.4.1` → `7.7.1`. That is also why CI passed: it did the same thing silently.

So it's 44 lines of diff noise that the next `pub get` undoes — and it happens to be the analyzer 7 → 8 jump that has its own open issues (#606, #611, #512). I'd drop it from the PR.

### 3. The win shrank considerably after the review fixes

The summary notes that each `ChatRoomNotifier` does a full history decrypt — and those are now eager for *every* session with a peer. What's left is skipping the `OrderNotifier` for five statuses older than 24 h.

In exchange, the startup loop now performs one `await messageStorage.getLatestMessageById()` **per session, sequentially**, including for every session that stays eager, where it is pure added cost.

Probably still a net win, but it's measurable and worth confirming before merge — especially since the title still promises more than the change now delivers.

### 4. Nits

- If one `messageId` ever held both an image and a file, the accounting breaks: I probed it and the image's 4 MB stays in the map while the counter drops to 1 MB — untracked, unevictable memory. In practice a message is one or the other, so this is robustness rather than a live bug.
- `_MediaCacheEntry.owner` is a strong reference held from a process-global `static` list. It's mitigated because both notifiers call `clearMediaCaches()` in `dispose()`, but a `WeakReference` wouldn't depend on dispose running — and this PR's own premise is that these notifiers live for the whole process.
- The bot asked for integration coverage of both P1 scenarios (restart with a canceled bond, chat on a terminal session). It wasn't added; the new tests are the pure decision table. Narrowing the predicate makes that largely moot, but noting it.

### Verdict

**Changes requested.**

Finding 1 is a real, reproducible bug with a one-word fix — an infinite download-and-decrypt loop is not something to ship in a PR whose whole point is reducing consumption. Finding 2 is cleanup, but cleanup that touches a toolchain decision tracked by its own issues.

The rest of the PR is solid. The responses to all three bot findings are substantive rather than cosmetic, and I verified the LRU fix working. With those two changes I'd approve.

grunch:
Feedback addressed in b0d2052.

**Catrya — `_MediaCacheEntry.owner` as a strong reference from a static list.** Now a `WeakReference<MediaCacheMixin>`. `_mediaTouch` and `clearMediaCaches()` compare against `owner.target`, and `_evict()` skips an already collected owner: its maps are gone with it, so only the accounting is dropped. The bound no longer depends on `dispose()` running.

**CodeRabbit — out-of-range timestamp.** `_messageTime` returns `null` past `8640000000000000` ms, so a corrupt stored value counts as an unknown age (eager) instead of throwing `RangeError` out of `isSettledOrderMessage` and aborting initialization. Regression test added; it reproduced the crash before the fix.

Not changed, deliberately:
- Integration coverage of the two Codex P1 scenarios: with `canceled` kept eager and the chat notifier eager for every peer session, the decision table already pins both, as noted in the previous review.
- Several mounted images summing over the budget: with the entry just touched never evicting itself, this is at worst one re-load per parent rebuild in that pathological case, not a loop. Happy to add a hysteresis if you want it in this PR.

`pubspec.lock` is untouched (the implicit `pub get` on test runs still rewrites it locally; reverted before committing). `flutter analyze` clean apart from the two pre-existing `containsSemantics` infos; `shared/` and `chat/` suites green (331 tests).

PR #732, opened 2026-09-13
fix(order): drop an invoice when its take cycle ends

### LINKED ISSUE #731: Stale escrow invoice shown as bond invoice after cancelling and re-taking the same order
## Description

After a taker cancels a take and then takes the same order again, the "anti-abuse bond" screen shows the **escrow invoice from the previous take** instead of a fresh bond invoice. That invoice was already canceled on the Lightning node, so paying it fails with `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` and the take cannot be completed.

## Steps to reproduce

1. Maker publishes a **buy** order with a fiat range (1–10 EUR, market price -10%).

Image

Image

2. Another user takes it as seller and pays the anti-abuse bond (5000 sats) — correct.

Image

3. Seller pays the escrow hold invoice for the full amount (16558 sats for 10 EUR) — correct.
4. 
Image

4. Order moves to `waiting-buyer-invoice` (buyer has not added their invoice yet).

Image

5. **Before** the buyer adds the invoice, the seller (taker) cancels the order.
6. Take the same order again, with any fiat amount.

## Expected

The bond screen shows a new bond invoice (~5000 sats, memo `mostro bond order_id=...`) generated for this new take.

## Actual

The bond screen ("Depósito anti-abuso") shows **16558 sats** — the full order amount — and the QR is the escrow bolt11 from the *previous* take. Its description reads:

```
Escrow amount Order #df88142b-468d-4e2a-b88c-7f34f3ab4dce: SELL BTC ...
```

Image

Paying it fails:

```
Failed to make payment
FAILURE_REASON_INCORRECT_PAYMENT_DETAILS
```

Image

The user is then stuck: the take cannot be paid for and cannot be completed.

## Why this is a client bug, not a mostrod bug

The memo is decisive. mostrod builds bond invoices with memo `mostro bond order_id={id}` (`mostro/src/app/bond/flow.rs`, `request_taker_bond`) and ships them in `Action::PayBondInvoice`. It never labels a bond invoice `Escrow amount Order #...`. So the bolt11 rendered on the bond screen cannot have come from a `PayBondInvoice` message — it is the `Action::PayInvoice` (escrow) payload from the first take, still held in local state. mostrod cancels that hold invoice during the taker-cancel flow (`cancel_order_by_taker_inner`), which is exactly why paying it returns `INCORRECT_PAYMENT_DETAILS`.

## Root cause

`OrderState` keeps `paymentRequest` forever, keyed by order id:

```dart
// lib/features/order/models/order_state.dart:168
} else {
  newPaymentRequest = paymentRequest; // Preserve existing
}
```

There is already a precedent for clearing per-take state when an order returns to the book — `orderReactivated` (status → `pending`) clears the taker reputation (`order_state.dart:158`, `:298`, `clearPeerReputation`) — but `paymentRequest` is not cleared alongside it.

Consumers then render whatever the last `PaymentRequest` was:

- `lib/features/order/screens/pay_bond_invoice_screen.dart:126-127` — `orderState.paymentRequest?.lnInvoice` / `?.order?.amount`
- `lib/features/trades/screens/trade_detail_screen.dart:371-380` — the "pay bond" button is shown whenever `paymentRequest != null`

## Suggested fix

Add a `clearPaymentRequest` flag to `OrderState.copyWith` and set it when the order is reactivated (`newStatus == Status.pending`), mirroring `clearPeerReputation`. Optionally also clear it when a new take starts, and have `PayBondInvoiceScreen` refuse to render an invoice whose action/phase doesn't match the current status (`waiting-taker-bond`) instead of falling back to the last known one.

## Note for the new web client (`MostroP2P/app`)

The new Flutter/Rust client does **not** have this bug today, for two independent reasons:

1. It doesn't implement bonds at all — `Action::PayBondInvoice` is mapped to a hard `Rejected { reason: "BondRequired" }` (`rust/src/api/orders.rs`, `classify_take_reply`), so there is no bond screen that could render a stale invoice.
2. Each take persists a **new** `TradeInfo` row (`id: Uuid::new_v4()`) carrying the `hold_invoice` from that take's own daemon reply, instead of a per-order field preserved across takes.

Two things worth keeping in mind there when bonds are implemented, since the same trap exists:

- `update_trade_fields` cannot clear `hold_invoice` — `None` means "leave untouched" (`rust/src/db/sqlite.rs`), so a canceled trade row keeps its dead bolt11. Harmless only because a retake inserts a new row.
- Rows for the same `order.id` are disambiguated by `ORDER BY started_at DESC` with second granularity; two takes of the same order within the same second leave the ordering undefined, and `firstOrNull` could pick the stale row (`lib/features/order/providers/trade_state_provider.dart`).

21Mill:
Fixes #731

## Problem

Take an order, pay the bond, pay the escrow, cancel as the taker before the buyer adds their invoice, then take the same order again: the "anti-abuse deposit" screen shows the **previous** cycle's escrow invoice (full order amount, memo `Escrow amount Order #…`). Mostro cancels that hold invoice in `cancel_order_by_taker_inner`, so paying it fails with `FAILURE_REASON_INCORRECT_PAYMENT_DETAILS` and the order ends up untakeable.

The memo is what identifies it as a client-side issue: mostrod builds bond invoices with memo `mostro bond order_id={id}` and ships them in `Action::PayBondInvoice`, so the bolt11 rendered on the bond screen never came from a bond message.

## Cause

Two things compound:

1. `OrderState` is keyed by order id and preserved `paymentRequest` indefinitely (`newPaymentRequest = paymentRequest; // Preserve existing`), so the escrow bolt11 outlived the take that produced it. The same reset already exists next to it for the taker reputation (`clearPeerReputation` when the order returns to `pending`).
2. A cancelled order outranks every waiting phase in `phaseRank`, so during `OrderNotifier.sync()`'s replay the first message of the *next* take looks like a late copy to `isStaleTransition` and the whole new cycle is dropped — leaving the previous cycle's invoice as the newest thing in state.

## Changes

- `OrderState.updateWith` drops `paymentRequest` on the statuses that end a take cycle (`endsTradeCycle`: pending, canceled, canceled-by-admin, cooperatively-canceled, expired). A message that carries its own `PaymentRequest` still wins, so the next cycle's invoice is never swallowed. `copyWith` gains `clearPaymentRequest`, mirroring `clearPeerReputation`.
- `OrderNotifier` resets its state when the user takes the order again (`_resetForNewTakeCycle`), and `sync()` restarts the replay from a clean state on the first message of a new cycle (`wouldRejectAsStale`) instead of dropping it.
- `PayBondInvoiceScreen` renders an invoice only while the order is in the bond phase and shows an explanatory empty state otherwise, so no entry point (notification card, trade detail button) can surface a dead bolt11. New key `bondInvoiceUnavailable` added to all six ARB files.

## Tests

New `test/features/order/models/order_state_retake_test.dart` — 9 tests covering the cycle boundaries, `endsTradeCycle`, `wouldRejectAsStale`, and the full bug sequence replayed the way `sync()` does it (bond → escrow → waiting-buyer-invoice → canceled → new bond ⇒ state ends on the new bond invoice).

`flutter analyze` is clean on the touched files. Locally 53 test files fail to load on an unmodified checkout too — `test/mocks.mocks.dart` is stale and `dart run build_runner build -d` doesn't compile against the analyzer version resolved here; CI regenerates them.

## Note for the new web client (`MostroP2P/app`)

It doesn't have this bug today: it has no bond flow (`PayBondInvoice` → `Rejected{BondRequired}`) and each take persists a new `TradeInfo` row with its own `hold_invoice`. Two things to keep in mind there when bonds land, since the same trap exists: `update_trade_fields` cannot clear `hold_invoice` (`None` = leave untouched), and rows for the same order id are disambiguated only by `started_at` at second granularity.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented outdated bond and escrow invoices from carrying over when an order is retaken.
  - Order replay now correctly distinguishes new take cycles from stale updates.
  - Bond invoices are shown only during the appropriate order phase.

- **User Experience**
  - Added clear pending, unavailable, and already-paid bond states with appropriate navigation.
  - Added translations for bond invoice messages and trade navigation in English, German, Spanish, French, Italian, and Portuguese.


lib/l10n/intl_fr.arb
```diff
@@ -843,6 +843,7 @@
     "copy": "Copier",
     "share": "Partager",
     "failedToShareInvoice": "Échec du partage de la facture. Veuillez essayer de copier à la place.",
+    "bondInvoiceUnavailable": "Cette facture de dépôt n'est plus valide. Reviens en arrière et reprends l'ordre pour en obtenir une nouvelle.",
```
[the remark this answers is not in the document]
21Mill: Fixed in a9ee9174 — thanks.

(Commit rewritten to re-sign it; content unchanged.)

lib/l10n/intl_pt.arb
```diff
@@ -843,6 +843,7 @@
     "copy": "Copiar",
     "share": "Compartilhar",
     "failedToShareInvoice": "Falha ao compartilhar a fatura. Por favor, tente copiar em vez disso.",
+    "bondInvoiceUnavailable": "Esta fatura de depósito já não é válida. Volta atrás e aceita a ordem novamente para obteres uma nova.",
```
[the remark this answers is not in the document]
21Mill: Fixed in a9ee9174 — thanks.

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/notifiers/order_notifier.dart
```diff
@@ -92,9 +92,25 @@ class OrderNotifier extends AbstractMostroNotifier {
...
-        if (message.action != Action.cantDo) {
-          currentState = currentState.updateWith(message);
+        if (message.action == Action.cantDo) continue;
+
+        // A cancelled order outranks every waiting phase, so to the
+        // stale-transition guard the first message of the *next* take of the
+        // same order id looks like a late copy and the whole new cycle would
+        // be dropped — leaving the previous cycle's invoice on screen (#731).
+        // Replay it from a clean slate instead.
+        if (OrderState.endsTradeCycle(currentState.status) &&
```
grunch: _🎯 Functional Correctness_ | _🔴 Critical_

**This restart bypasses the stale-transition guard on every terminal order, not only on retakes.**

`endsTradeCycle(current) && wouldRejectAsStale(message)` is true for *every* message the guard exists to drop once an order is canceled, canceled-by-admin or expired.

The replay sorts by `eventCreatedAt`, which only has one-second resolution, and uses the local receive time as tie-break. Relays replay newest-first, so a late copy from the same second as the `canceled` sorts **after** it. Before this PR, `updateWith` dropped that copy. Now the state is rebuilt from `pending` and the copy is applied, so a canceled trade comes back as `active` or `waiting-payment`.

`order_state_late_setup_message_test.dart` (*late copies on terminal orders*) guards against exactly this at model level, and it now comes back at notifier level. I verified it with the real `sync()`: `canceled`@3000 followed by `hold-invoice-payment-accepted`@3000 (received later) ends on `active`. The same history passes on `main`.

The message alone can't tell a new take from a late copy, so the restart needs positive evidence of a new cycle:

1. the message is a **cycle-opening** action (`pay-bond-invoice`, `pay-invoice`, `add-invoice` outside a payout retry, `waiting-seller-to-pay`, `waiting-buyer-invoice`), **and**
2. its `eventCreatedAt` is **strictly greater** than that of the message that ended the cycle. A same-second tie is not a new take, because Mostro can't cancel and accept a new take in the same second.

```dart
int? cycleEndedAt;
for (final message in messages) {
  if (message.action == Action.cantDo) continue;
  final eventTime = message.eventCreatedAt ?? message.timestamp ?? 0;
  if (OrderState.endsTradeCycle(currentState.status) &&
      OrderState.opensTakeCycle(message.action) &&
      cycleEndedAt != null && eventTime > cycleEndedAt &&
      currentState.wouldRejectAsStale(message)) {
    currentState = OrderState(status: Status.pending, action: Action.newOrder, order: currentState.order);
    cycleEndedAt = null;
  }
  final before = currentState.status;
  currentState = currentState.updateWith(message);
  if (!OrderState.endsTradeCycle(before) && OrderState.endsTradeCycle(currentState.status)) {
    cycleEndedAt = eventTime;
  }
}
```

Please add the tie case from the review body as a notifier-level regression test.

<details><summary>🤖 Prompt for AI Agents</summary>

```
In lib/features/order/notifiers/order_notifier.dart sync(), restrict the new-cycle reset to cycle-opening actions whose eventCreatedAt is strictly greater than the event time of the message that moved the state into an endsTradeCycle status. Add a test in test/features/order/notifiers/ using the order_notifier_replay_order_test.dart harness: canceled@3000 then hold-invoice-payment-accepted@3000 with a later local timestamp must stay canceled.
```
</details>

lib/features/order/models/order_state.dart
```diff
@@ -109,6 +112,19 @@ class OrderState {
...
+  /// to the book or ended for good.
+  ///
+  /// Everything Mostro issued for that cycle — bond and escrow hold invoices
+  /// above all — is cancelled node-side when this happens, so no payload from
+  /// it may survive into the next take of the same order id.
+  static bool endsTradeCycle(Status status) =>
+      status == Status.pending ||
+      status == Status.canceled ||
+      status == Status.canceledByAdmin ||
+      status == Status.cooperativelyCanceled ||
```
grunch: _🎯 Functional Correctness_ | _🟠 Major_

**`cooperativelyCanceled` does not end a take cycle.**

In this codebase it is the *pending* cooperative cancel: every `cooperative-cancel-initiated-*` action maps to it. The `phaseRank` doc further down says so: *"the trade can still reach fiat-sent while the cancel is pending"*. The real end is `cooperative-cancel-accepted`, which maps to `Status.canceled` and is already covered.

Listing it here causes two problems:

- `updateWith` drops `paymentRequest` in the middle of a live trade.
- Combined with the `sync()` restart, any late setup message after the cancel request resets the state to `pending`. That loses:
  - `fiatWasSent`, which drives the *FiatSent* vs *NoFiat* remap;
  - `peer`;
  - `dispute`. With `dispute == null`, `rejectsAdminDisputeMessage` rejects every later `admin-settled` / `admin-canceled`, so the dispute can no longer be resolved on the device.

I verified the `fiatWasSent` loss: `fiat-sent-ok`@2000 → `cooperative-cancel-initiated-by-peer`@3000 → late `buyer-took-order`@3000 ends on `active` with `fiatWasSent == false`. The same history passes on `main`.

Remove this line, and move `Status.cooperativelyCanceled` to the `isFalse` list in the `endsTradeCycle` test:

```suggestion
```

lib/features/order/screens/pay_bond_invoice_screen.dart
```diff
@@ -134,6 +143,44 @@ class PayBondInvoiceScreen extends ConsumerWidget {
         false;
     final explanation = isMakerBond ? s.bondExplanationMaker : s.bondExplanation;
 
+    if (lnInvoice.isEmpty) {
```
grunch, line 147: _🎯 UX / Functional Correctness_ | _🟠 Major_

**`lnInvoice.isEmpty` covers three different situations, and tells the user the invoice is dead in all of them.**

1. **Not loaded yet (maker bond).** `AddOrderNotifier._handleMakerBondInvoice` updates *its own* state, then calls `go('/pay_bond/<orderId>')`. If `orderNotifierProvider(orderId)` doesn't exist yet, the screen creates it. Its initial state is `pending/new-order` with no invoice until the async `sync()` completes, so the first frames show *"This bond invoice is no longer valid… take the order again"* with a CLOSE button to `/`. `bondExplanationMaker` warns that leaving this screen means the order won't be created, so one tap abandons a valid maker bond.
2. **Bond already paid.** After `bond-invoice-accepted` the flow moves on (`pay-invoice`, `add-invoice`, `waiting-*`). `handleEvent` only navigates away when the event is less than 60 s old (`isRecent`). A user who returns from the wallet late, or reopens this screen from the notification card or the back stack, is told to *take the order again* while their trade is progressing.
3. **Cycle really ended** (canceled, or pending again). This is the only case the copy describes.

Please branch on the order state rather than on the invoice string:
- no invoice while the order is still `pending` / `waiting-taker-bond` → loading indicator, no destructive action;
- status past the bond phase → redirect to `/trade_detail/$orderId`, or at least show a neutral message;
- `endsTradeCycle(status)` → the current empty state.

`pay_bond_invoice_screen.dart` has no tests today, so please add one widget test per branch.

lib/features/order/notifiers/order_notifier.dart
```diff
@@ -156,6 +187,9 @@ class OrderNotifier extends AbstractMostroNotifier {
       // it can't delete the session we just created (retake within 60s).
       AbstractMostroNotifier.clearBondCancelDeletion(orderId);
 
+      // Same reason, for the state this notifier still holds from that cycle.
+      _resetForNewTakeCycle();
```
grunch: _🎯 Robustness_ | _🟡 Minor_

**The reset runs before the take is published or acknowledged.**

- If `mostroService.takeSellOrder` throws, or Mostro answers `cant-do` (order already taken, `pending_order_exists`, out of range), the in-memory state becomes `pending/new-order` while storage still holds the previous cycle's history. Trade detail and the order list show a status that no stored message supports until the next `sync()`.
- `sync()` only reads `state` *after* `await storage.getAllMessagesForOrderId`. If a sync is running when the user retakes (the constructor sync, or the `onAdminResolutionRejected` replay chain), it overwrites the reset with the replayed `canceled` state. The live `subscribe()` path has no new-cycle logic, so the first message of the new take is then dropped as stale. That is the original #731 symptom.

Options:
- reset only the in-memory payloads (`clearPaymentRequest`, peer, reputation) and keep the status;
- move the reset to the first cycle-opening response of the new take;
- or, if the reset stays here, bump a `_cycleGeneration` counter that `sync()` checks before assigning `state`.

test/features/order/models/order_state_retake_test.dart
```diff
@@ -0,0 +1,185 @@
...
+        bond.wouldRejectAsStale(invoiceMessage(Action.payInvoice, 'lnbcescrow')),
+        isFalse,
+      );
+    });
+  });
+
+  group('replaying the persisted history of a retaken order', () {
+    // Mirrors the loop in OrderNotifier.sync: a message that only looks stale
+    // because the previous cycle ended restarts the replay.
+    OrderState replay(List<MostroMessage> messages) {
```
grunch: _🧪 Test Quality_ | _🟡 Minor_

**This helper copies the `sync()` loop instead of testing it.**

If `OrderNotifier.sync()` changes (for example with the fix proposed for the restart condition), these tests keep passing against the copy. `order_notifier_replay_order_test.dart` already has a working harness (in-memory Sembast, `_SyncOnlyOrderNotifier`). Please move the replay scenarios there so they run the real `sync()`, and add the negative cases:
- a same-second late copy after `canceled`, `canceled-by-admin` or `expired` must **not** reopen the order;
- a late setup message while `cooperatively-canceled` must keep `fiatWasSent`, `dispute` and `peer`;
- `_resetForNewTakeCycle` through `takeSellOrder` / `takeBuyOrder`, including a `cant-do` response.

Also, no message here sets `eventCreatedAt`, so ordering by event time is never exercised.

test/features/order/models/order_state_retake_test.dart
```diff
@@ -0,0 +1,185 @@
...
+    int amount = 16558,
+  }) =>
+      MostroMessage<PaymentRequest>(
+        id: 'order-1',
+        action: action,
+        payload: PaymentRequest(
+          order: orderPayload(),
+          lnInvoice: bolt11,
+        ),
+        timestamp: amount,
```
grunch: _🧪 Test Quality_ | _🔵 Nit_

**`amount` is written into `timestamp`.**

The `amount` parameter is never used as an amount. It sets `MostroMessage.timestamp` to `16558`, while the payload amount is hard-coded in `orderPayload()`. This looks like a copy/paste slip. It's harmless now, but misleading once a test relies on `timestamp` for ordering or the recency gate. Drop the parameter or pass it to `Order.amount`, and set `eventCreatedAt` explicitly wherever order matters.

lib/features/order/models/order_state.dart
```diff
@@ -460,6 +486,19 @@ class OrderState {
...
 
+  /// Whether [message] would be dropped by the stale-transition guard.
+  ///
+  /// Lets callers replaying persisted history tell a genuinely late copy from
+  /// the first message of a *new* take cycle: after a cancel, every later
+  /// message looks backwards to the guard, because a cancelled order outranks
+  /// every waiting phase. Combined with [endsTradeCycle] on the current
+  /// status, that is the signal to start the replay over instead of dropping
+  /// the rest of the history (#731).
+  bool wouldRejectAsStale(MostroMessage message) => isStaleTransition(
```
grunch: _♻️ Maintainability_ | _🔵 Nit_

**`wouldRejectAsStale` derives the status on its own.**

`updateWith` computes the status from `effectiveAction`, after the cooperative-cancel remap, which depends on `fiatWasSent`. This predicate uses the raw `message.action` instead. Today every remapped variant lands on `cooperativelyCanceled`, so the two agree, but they can drift apart without anyone noticing.

Extract a private `({Action action, Status status}) _transitionFor(MostroMessage)` and call it from both `updateWith` and `wouldRejectAsStale`. The predicate then always matches what `updateWith` would do.

lib/features/order/screens/pay_bond_invoice_screen.dart
```diff
@@ -134,6 +143,44 @@ class PayBondInvoiceScreen extends ConsumerWidget {
...
+                textAlign: TextAlign.center,
+                style: const TextStyle(
+                  color: AppTheme.cream1,
+                  fontSize: 15,
+                  height: 1.4,
+                ),
+              ),
+              const SizedBox(height: 24),
+              ElevatedButton(
+                onPressed: () => context.go('/'),
```
grunch: _🎨 UX_ | _🔵 Nit_

**The copy and the button disagree.**

`bondInvoiceUnavailable` tells the user to *"Go back and take the order again"*. The only button is **CLOSE**, and it calls `context.go('/')`, which opens the order book root instead of going back. Either pop (`context.pop()` when `canPop`, else `go('/')`), or reword the copy to match ("Return to the order book and take the order again…").

lib/l10n/intl_pt.arb
```diff
@@ -843,6 +843,7 @@
     "copy": "Copiar",
     "share": "Compartilhar",
     "failedToShareInvoice": "Falha ao compartilhar a fatura. Por favor, tente copiar em vez disso.",
+    "bondInvoiceUnavailable": "Esta fatura de depósito já não é válida. Volte atrás e aceite a ordem novamente para obter uma nova.",
```
grunch: _🌐 Localization_ | _🔵 Nit_

**Still reads as pt-PT.**

421c3a4b fixed the verb forms, but *"já não é válida"* and *"Volte atrás"* are still European Portuguese. The rest of this file is pt-BR (`você`), which would say:

```suggestion
    "bondInvoiceUnavailable": "Esta fatura de depósito não é mais válida. Volte e aceite a ordem novamente para obter uma nova.",
```

grunch (changes requested):
## Review (strict pass)

The diagnosis in #731 is right. The `updateWith` half of the fix is sound: it drops `paymentRequest` when a take cycle ends, and an incoming `PaymentRequest` always wins. The `sync()` restart, though, reopens a hole that #723 closed. `endsTradeCycle` also counts a status that is not terminal as a cycle end. I reproduced both against the real `OrderNotifier.sync()`, using the harness from `order_notifier_replay_order_test.dart`. Both tests **pass on `main` and fail on this branch**:

| Scenario (persisted history, sorted by `compareByEventTime`) | `main` | this PR |
|---|---|---|
| `waiting-seller-to-pay`@1000 → `hold-invoice-payment-accepted`@2000 → `canceled`@3000 → same-second late copy of `hold-invoice-payment-accepted`@3000 (received later) | `canceled` ✅ | **`active`** ❌ |
| `buyer-took-order`@1000 → `fiat-sent-ok`@2000 → `cooperative-cancel-initiated-by-peer`@3000 → same-second late `buyer-took-order`@3000 | `cooperatively-canceled`, `fiatWasSent=true` ✅ | **`active`**, `fiatWasSent` reset ❌ |

<details><summary>Reproduction test</summary>

```dart
// Same setUp/_SyncOnlyOrderNotifier as order_notifier_replay_order_test.dart;
// message(...) additionally takes `timestamp` (local receive time, tie-break).
test('same-second late copy after canceled must not reopen', () async {
  final storage = container.read(mostroStorageProvider);
  await storage.addMessage('a', message(Action.waitingSellerToPay, Status.waitingPayment, eventCreatedAt: 1000, timestamp: 1));
  await storage.addMessage('b', message(Action.holdInvoicePaymentAccepted, Status.active, eventCreatedAt: 2000, timestamp: 2));
  await storage.addMessage('c', message(Action.canceled, Status.canceled, eventCreatedAt: 3000, timestamp: 3));
  await storage.addMessage('d', message(Action.holdInvoicePaymentAccepted, Status.active, eventCreatedAt: 3000, timestamp: 4));
  await container.read(orderNotifierProvider(orderId).notifier).sync();
  expect(container.read(orderNotifierProvider(orderId)).status, Status.canceled);
});

test('late copy during pending cooperative cancel keeps fiatWasSent', () async {
  final storage = container.read(mostroStorageProvider);
  await storage.addMessage('a', message(Action.buyerTookOrder, Status.active, eventCreatedAt: 1000, timestamp: 1));
  await storage.addMessage('b', message(Action.fiatSentOk, Status.fiatSent, eventCreatedAt: 2000, timestamp: 2));
  await storage.addMessage('c', message(Action.cooperativeCancelInitiatedByPeer, Status.active, eventCreatedAt: 3000, timestamp: 3));
  await storage.addMessage('d', message(Action.buyerTookOrder, Status.active, eventCreatedAt: 3000, timestamp: 4));
  await container.read(orderNotifierProvider(orderId).notifier).sync();
  final state = container.read(orderNotifierProvider(orderId));
  expect(state.status, Status.cooperativelyCanceled);
  expect(state.fiatWasSent, isTrue);
});
```
</details>

### Findings

| # | Severity | Where | Summary |
|---|---|---|---|
| 1 | 🔴 Critical | `order_notifier.dart` `sync()` | The retake restart fires for **any** message the stale guard rejects, so a same-second late copy reopens a terminal trade (verified) |
| 2 | 🟠 Major | `order_state.dart` `endsTradeCycle` | `cooperativelyCanceled` is a pending cancel, and the trade can go on. Counting it as a cycle end wipes `fiatWasSent`, `peer` and `dispute` during a replay (verified for `fiatWasSent`) |
| 3 | 🟠 Major | `pay_bond_invoice_screen.dart` | The empty state treats "invoice not loaded yet" and "bond already paid" as "invoice expired", and its only action abandons the flow |
| 4 | 🟡 Minor | `order_notifier.dart` `_resetForNewTakeCycle` | State is wiped before the take is published or acknowledged, so a failed publish or a `cant-do` leaves it out of sync with storage, and it can race an in-flight `sync()` |
| 5 | 🟡 Minor | `order_state_retake_test.dart` | The replay test copies the `sync()` loop instead of running it, and the negative cases are missing |
| 6 | 🔵 Nit | `order_state.dart` `wouldRejectAsStale` | Repeats the status derivation from `updateWith` without its cooperative-cancel remap |
| 7 | 🔵 Nit | l10n / button | The copy says "Go back" but the button is CLOSE → `/`, and the pt string still reads as pt-PT |

Checks run locally on `421c3a4b` with Flutter 3.35.7, mocks regenerated:
- `flutter analyze lib/features/order test/features/order` reports no issues in the PR files.
- All 160 tests under `test/features/order` pass, including the 9 new ones.
- The two reproduction tests above are the only failures, and both are regressions.

**Verdict: request changes.** Items 1–3 should be fixed before merge.

lib/features/order/notifiers/order_notifier.dart
21Mill: Fixed in e4c94b3f. The restart now needs both halves of the evidence you describe — a cycle-opening action (OrderState.opensTakeCycle) and an event time strictly later than the message that ended the previous cycle — and the rule lives in one place, AbstractMostroNotifier.applyToCycle, so the live stream gets it too. Your same-second repro is a test now: "same-second copy of a setup message after canceled" in test/features/order/notifiers/order_notifier_retake_cycle_test.dart, plus the mid-trade variant.

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/models/order_state.dart
21Mill: Fixed in e4c94b3f — you are right, and the code says so itself ("Actions that should set status to cooperatively canceled (pending cancellation)") and its action table still offers dispute/DM. Removed from endsTradeCycle, moved to the isFalse list, and your fiatWasSent repro is now a notifier test.

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/screens/pay_bond_invoice_screen.dart
21Mill: Fixed in 2bed2518. Four branches now: invoice in the bond phase, spinner with no destructive action while it loads (the maker case you traced through AddOrderNotifier), a pointer to /trade_detail once the bond is paid, and the expired copy only when endsTradeCycle holds. One widget test per branch in test/features/order/screens/pay_bond_invoice_screen_test.dart — the screen had none.

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/notifiers/order_notifier.dart
21Mill: Fixed in e4c94b3f by removing _resetForNewTakeCycle entirely. With applyToCycle on the live path, nothing needs to change before the take is acknowledged: a cant-do or a failed publish leaves the state untouched, and there is no reset for an in-flight sync() to overwrite.

(Commit rewritten to re-sign it; content unchanged.)

test/features/order/models/order_state_retake_test.dart
21Mill: Fixed in e4c94b3f. The replay scenarios moved to test/features/order/notifiers/order_notifier_retake_cycle_test.dart and run the real sync() with the harness from order_notifier_replay_order_test.dart; every message carries eventCreatedAt and timestamp. The negative cases are there: same-second late copy after canceled (setup action and mid-trade action) and the pending cooperative cancel keeping fiatWasSent. One note: the harness imports test/mocks.mocks.dart only for MockSessionStorage, and those generated mocks do not compile against the analyzer version resolving here, so the new file uses a local no-op SessionStorage instead and runs. canceled-by-admin is not driven from a test because an admin-* message is itself dropped without a tracked dispute, which would prove nothing about the cycle rule.

(Commit rewritten to re-sign it; content unchanged.)

test/features/order/models/order_state_retake_test.dart
21Mill: Fixed in e4c94b3f — copy/paste slip, the parameter is gone and the notifier tests set eventCreatedAt and timestamp explicitly wherever order matters.

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/models/order_state.dart
21Mill: Fixed in e4c94b3f. Both go through a new _transitionFor(message, {fiatSent}) that does the cooperative-cancel remap and the status derivation, so wouldRejectAsStale always matches what updateWith would apply.

(Commit rewritten to re-sign it; content unchanged.)

lib/l10n/intl_pt.arb
21Mill: Fixed in 2bed2518: "não é mais válida" instead of "já não é válida".

(Commit rewritten to re-sign it; content unchanged.)

lib/features/order/notifiers/abstract_mostro_notifier.dart
```diff
@@ -45,6 +45,53 @@ class AbstractMostroNotifier extends StateNotifier<OrderState> {
...
+        order: current.order,
+      );
+    }
+
+    final next = current.updateWith(message);
+
+    // Recorded from the resulting status rather than from the transition: a
+    // replay starts from the current state, so the message that ended the
+    // cycle can be applied onto an already-ended one.
+    cycleEndedAt = OrderState.endsTradeCycle(next.status) ? eventTime : null;
```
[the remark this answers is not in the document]
21Mill: Fixed in 05266087 — confirmed before fixing: from waitingTakerBond the guard treats a stale `canceled` (rank 6) as a forward move, so it applied and cleared the new cycle's paymentRequest. applyToCycle now tracks cycleStartedAt alongside cycleEndedAt and drops anything older, and subscribe() consults the same predicate before any side effect. Regression test: "a cancel from the previous cycle cannot void the new invoice" in test/features/order/notifiers/order_notifier_retake_cycle_test.dart.

lib/features/order/notifiers/abstract_mostro_notifier.dart
```diff
@@ -134,7 +181,7 @@ class AbstractMostroNotifier extends StateNotifier<OrderState> {
                   _userInitiatedCancels.remove(orderId);
 
               if (mounted) {
-                state = state.updateWith(msg);
+                state = applyToCycle(state, msg);
```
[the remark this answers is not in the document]
21Mill: Partly fixed in 05266087, and the rest is out of scope here. The cycle-boundary case you describe is now gated exactly as you suggest: precedesActiveCycle is consulted in subscribe() before the timeout cancellation, the cancel-flag consumption, handleEvent, navigation and session persistence — the same shape as the existing rejectsAdminDisputeMessage gate. The general case (any message updateWith drops as a stale transition still running its side effects) predates this PR: on main `state = state.updateWith(msg)` silently returns `this` and handleEvent runs anyway. Widening the gate to every stale drop changes behaviour well beyond #731, so it belongs in its own change rather than in a bugfix PR that is already under review.

test/features/order/notifiers/order_notifier_retake_cycle_test.dart
```diff
@@ -0,0 +1,330 @@
...
+  Future<void> persist(List<(String, MostroMessage)> history) async {
+    final storage = container.read(mostroStorageProvider);
+    for (final (key, msg) in history) {
+      await storage.addMessage(key, msg);
+    }
+  }
+
+  Future<OrderStateSnapshot> syncedState() async {
+    await container.read(orderNotifierProvider(orderId).notifier).sync();
+    final state = container.read(orderNotifierProvider(orderId));
```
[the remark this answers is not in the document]
21Mill: Fixed in 05266087. You are right that the constructor's unawaited sync() made the second call a no-op resync request; the helper now lets the constructor pass drain before running its own sync() on a quiet notifier, so the assertions always read the hydrated state. (The same pattern exists in the pre-existing order_notifier_replay_order_test.dart harness this file was modelled on.)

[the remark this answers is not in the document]
21Mill: You are right — the ten event-loop turns narrowed the window instead of closing it; the storage read can still be pending on the tenth. Fixed properly in eb677e22 with your second option: _SyncOnlyOrderNotifier now swallows the constructor's unawaited sync(), so the pass the helper awaits is the only one that runs and hydration is deterministic. The helper is back to a single `await notifier.sync()`. The pre-existing order_notifier_replay_order_test.dart harness still has the original timing dependency; I left it alone since it is outside this PR, but it is worth the same treatment.

